PR #1, opened 2018-04-23
Feature/initial commit

linda-mccall:
db schema, entity relationship models, service classes, REST controllers, JWT authentication


pom.xml
```diff
@@ -0,0 +1,129 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.microsoft.sqlserver</groupId>
```
BrownMark: Is this required?

pom.xml
```diff
@@ -0,0 +1,129 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.unosquare</groupId>
+    <artifactId>hr-manager-service</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <name>hr-manager-service</name>
+    <description>Demo project for Spring Boot</description>
```
BrownMark: Can we keep it to admin core for now? Just to fall in line with the rest of unosquare's internal conventions

src/main/java/com/unosquare/adminCore/ApplicationStartup.java
```diff
@@ -0,0 +1,45 @@
...
+     */
+    @Override
+    public void onApplicationEvent(final ApplicationReadyEvent event) {
+
+        employeeUserRoleService.save(EmployeeUserRoles.TeamLeader.getRole());
+        employeeUserRoleService.save(EmployeeUserRoles.SystemAdmin.getRole());
+        employeeUserRoleService.save(EmployeeUserRoles.User.getRole());
+
+        if (employeeService.findByEmail("superuser@unosquare.com") == null) {
+            SignUpRequest signUpRequest = new SignUpRequest("super", "user",
```
BrownMark: This is fine for now, we'll be switching over to an SSO provider in the future anyway

src/main/java/com/unosquare/adminCore/HrManagerServiceApplication.java
```diff
@@ -0,0 +1,23 @@
...
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.builder.SpringApplicationBuilder;
+import org.springframework.boot.web.servlet.support.SpringBootServletInitializer;
+import springfox.documentation.swagger2.annotations.EnableSwagger2;
+
+@SpringBootApplication
+@EnableSwagger2
+public class HrManagerServiceApplication extends SpringBootServletInitializer {
```
BrownMark: Can we rename this to something inline with admin core?

src/main/java/com/unosquare/adminCore/entity/Holiday.java
```diff
@@ -0,0 +1,59 @@
...
+        this.lastModified = LocalDate.now();
+        this.dateCreated = LocalDate.now();
+    }
+
+    public Holiday(LocalDate startDate, LocalDate endDate, short holidayLength, Employee employee, String status) {
+        this.startDate = startDate;
+        this.endDate = endDate;
+        this.holidayLength = holidayLength;
+        this.employee = employee;
+        this.status = "new";
```
BrownMark: Suspect a copy paste error here.

src/main/java/com/unosquare/adminCore/enums/EmployeeUserRoles.java
```diff
@@ -0,0 +1,28 @@
+package com.unosquare.adminCore.enums;
+
+import com.unosquare.adminCore.entity.EmployeeUserRole;
+
+public enum EmployeeUserRoles {
```
BrownMark: Can we refactor to EmployeeRoles, with Employee representing the user in the system I think this will cause confusion.

BrownMark:
We need to add the following types as data models for referential integrity:

1. Country
2. Client Status
3. Contract Status
4. Holiday Status
5. Security roles

The enums should be updated with the IDs for these fields (we'll need a migration strategy for this  (e.g. https://flywaydb.org/getstarted/firststeps/maven)- you'll need to set the IDs explicitly, this would be done with set identity_insert on sql server and is a very manual process with postgres).

BrownMark:
We need to get away from the dal data models being passed in the spring controllers. It ties our hands, can result in lazy loading serialization problems and can result in a lot of meaningless data being sent across the wire.

We should be creating DTOs for now, i'd take a look at model mapper or similar to save some time: http://modelmapper.org/getting-started/

Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM openjdk:8
```
BrownMark: Openjdk is pretty large, this definitely lends itself well to a multi stage dockerfile

src/main/java/com/unosquare/adminCore/ApplicationStartup.java
```diff
@@ -0,0 +1,45 @@
...
+
+@Component
+public class ApplicationStartup
+        implements ApplicationListener<ApplicationReadyEvent> {
+
+    @Autowired
+    EmployeeService employeeService;
+
+    @Autowired
+    EmployeeUserRoleService employeeUserRoleService;
```
BrownMark: Yeah lets change this to EmployeeRole, it sounds like it's joining employee, user and roles

src/main/java/com/unosquare/adminCore/entity/Contract.java
```diff
@@ -0,0 +1,38 @@
...
+    @JoinColumn(name = "clientId", referencedColumnName = "clientId", insertable = false, updatable = false)
+    private Client client;
+
+    private String status;
+
+    public Contract() {
+
+    }
+
+    public Contract(ContractPK id, String status) {
```
BrownMark: the status needs to be an integer here with a separate table for referential integrity.

src/main/java/com/unosquare/adminCore/entity/Holiday.java
```diff
@@ -0,0 +1,59 @@
...
+    public Holiday() {
+
+    }
+
+    public Holiday(LocalDate startDate, LocalDate endDate, short holidayLength, Employee employee) {
+        this.startDate = startDate;
+        this.endDate = endDate;
+        this.holidayLength = holidayLength;
+        this.employee = employee;
+        this.status = "new";
```
BrownMark: I don't think the holiday status enum has a value of "new", again there should really be another table for the ref data.

src/main/java/com/unosquare/adminCore/controller/ContractController.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.web.bind.annotation.*;
+
+import java.util.List;
+
+@RestController
+@RequestMapping("/contracts")
+public class ContractController {
+
+    @Autowired
+    ContractService contractService;
```
BrownMark: I'll add a PR wide comment, but we really shouldn't be serializing data models for our request. Separate out into DTOs for requests to keep the payload to what you need and reduce the risk of lazy loading (here's a quick overview https://howtoprogramwithjava.com/hibernate-eager-vs-lazy-fetch-type/ )

src/main/java/com/unosquare/adminCore/entity/Holiday.java
```diff
@@ -0,0 +1,59 @@
...
+public class Holiday {
+
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
+    private int holidayId;
+
+    private LocalDate startDate;
+    private LocalDate endDate;
+
+    private short holidayLength;
```
BrownMark: Is holiday length needed? Having something like this places a requirement on having to keep it up to date and as we add more criteria in it will require tasks to keep in current.

I'd keep it to the date range only for now, performance wouldn't be an issue.

BrownMark (changes requested):
Hey Linda, some changes, nothing too major though. Largely looks great

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM maven:3.5-jdk-8 as BUILD
+ 
+COPY src /usr/src/myapp/src
+COPY src/main/resources/application.properties.docker /usr/src/myapp/src/main/resources/application.properties
```
BrownMark: can we update the target directory to not contain myapp.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY pom.xml /usr/src/myapp
+RUN mvn -f /usr/src/myapp/pom.xml clean package
+ 
+FROM openjdk:8-jre-alpine
+ 
+COPY --from=BUILD /usr/src/myapp/target/admin-core-0.0.1-SNAPSHOT.jar /opt/jboss/wildfly/standalone/deployments/admin-core.jar
```
BrownMark: Can we change the target directory to not include jboss wildfly etc, this was copied and pasted from a sample

pom.xml
```diff
@@ -0,0 +1,145 @@
...
+    <version>0.0.1-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <name>admin-core</name>
+    <description>Administration tool for booking and approving holidays</description>
+
+    <parent>
+        <groupId>org.springframework.boot</groupId>
+        <artifactId>spring-boot-starter-parent</artifactId>
+        <version>2.0.0.RELEASE</version>
```
BrownMark: Is it worth using 2.1 here? I seem to remember seeing news on flaws present in 2.0

src/main/java/com/unosquare/adminCore/controller/ClientController.java
```diff
@@ -0,0 +1,88 @@
...
+
+@CrossOrigin(origins = "*", allowCredentials = "true", allowedHeaders = "*")
+@RestController
+@RequestMapping("/clients")
+public class ClientController {
+
+    @Autowired
+    ClientService clientService;
+
+    private ModelMapper modelMapper = new ModelMapper();
```
BrownMark: We should probably inject this. Gives us flexibility in controlling scope etc

src/main/java/com/unosquare/adminCore/controller/ClientController.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @GetMapping(value = "/findByClientStatus/{clientStatusId}", produces = MediaType.APPLICATION_JSON_VALUE)
+    @ResponseStatus(HttpStatus.OK)
+    public List<ClientDto> findByClientStatus(@PathVariable("clientStatusId") short clientStatusId) {
+       return mapClientsToDtos(clientService.findByClientStatus(ClientStatus.fromId(clientStatusId)));
+    }
+
+    private List<ClientDto> mapClientsToDtos(List<Client> clients){
+        return clients.stream().map(client -> modelMapper.map(client, ClientDto.class)).collect(Collectors.toList());
```
BrownMark: I'm pretty sure model mapper supports list mapping e.g. modelMapper.map(clients,  ArrayList<Client>.class)

BrownMark (changes requested):
Just some minor final changes

BrownMark:
Last note really, if you get sonar lint installed can you run a quick scan in the app and fix any errors/warnings

BrownMark (approved):
Looks good Linda, just a few small issues but we can iron those out later

PR #3, opened 2018-05-18
Modified Makefile to be able to build both docker images

gdeak-unosquare:
Updated the Makefile


BrownMark (approved):
Looks good Gabe, good work man

PR #7, opened 2018-05-21
ESLint Integration

bgribz:
- Adds ESLint 
- Adds class properties babel transform
- Adds linting npm scripts

I've just fired in an ESLint config I've used in the past that I found to be good - we can discuss whatever rules we want to use, or maybe we want to use a prebuilt one. 

I've only added the class properties transform to babel, maybe we should look into using a preset instead?


BrownMark (approved):
Looks grand

PR #8, opened 2018-05-22
Feature/4.5 env variables

fraser-uno:
Added dotenv-webpack npm package to the application. 

Makes sense going forward when integrating with AWS and different environments.

I've added the .env to the .gitignore, probably not necessary in this project but just going with standards.
So you will need to create your own .env file in the web-app/ directory

I've added the Set up involed to the Wiki


BrownMark (approved):
Looks good

PR #10, opened 2018-05-22
Feature/axios (4.3)

William-Unosquare:


BrownMark (approved):
Looks good. Like you say there will need to be a chore to update the current requests, but the util is in for now

PR #11, opened 2018-05-22
Add stage-2 preset to babel

William-Unosquare:
Adds the ability to use stage 2 features like the spread operator.  Also removes the plugin `transform-class-properties` as this is included in the stage 2 preset.


BrownMark (approved):
Looks good Will

PR #13, opened 2018-05-22
Revise auth/routes

bgribz:
A lot of the changes are linting fixes, so it's probably best to pull this branch down and have a nosey.

This is just a quick first attempt at tidying up/reorganising the routing/auth logic. I'm not entirely sold on it - I feel like including the Header in the auth HOC isn't a particularly nice solution but I couldn't figure out a nice way of giving the Header access to the router's history so that it can perform navigation. 

Any thoughts?


ghost:
Might be one I'll pull down and have a look at with you. Haven't got time today but can do tomorrow AM.

web-app/src/pages/Login/LoginService.js
```diff
@@ -44,7 +44,7 @@ export default class LoginService {
   }
 
   setUserEmail(email) {
-    localStorage.setItem('user_email', email);   
+    localStorage.setItem('user_email', email);
```
ghost: Not sure I like this.....I guess this is where we could be setting email in the state....but it would be overkill for the little amount of state we have.

bgribz:
I've removed the `Header` from the auth HOC in favour of a `Layout` component we can wrap any pages in. I think it's a wee bit cleaner this way.

web-app/src/pages/Login/LoginService.js
peter-drysdale: This is something I added because we had no way of getting the user ID save for getting the profile with the API call that takes parameters 'firstname' and 'surname', which could only at the time be retrieved from the successful email. If we're changing login from JWT to base64 encode with user ID in payload then this method can go.

ghost: Cool makes sense!

web-app/src/components/TakenLeave/index.jsx
```diff
@@ -1,29 +1,35 @@
...
-                      {formatDate(holiday.date)}
-                  </span>
-              })
-          }
-      </div>
-    );
+  return (
+    <div>
+      {
+        props.takenHolidays.map(holiday => {
```
ghost: What is props here? What exists in that object? Would be nice to spread it then take what we need and use a better naming convention.

bgribz: It's just whatever the wrapping container is passing down. Although I think we need to discuss where containers are kept/what needs a container, so should probably address this in a different PR.

ghost: hmmm might be better to get const { takenHolidays } = this.props; 

I'm doing this blind so not sure if it'll work with your setup but this seems more readable.

William-Unosquare:
@bgribz I'm not sure if you solved this yesterday, but you can give components access to the router's history with `withRouter`, unless was there a reason you couldn't do this?

https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md

BrownMark (approved):
Looks good Barry.

PR #17, opened 2018-05-25
Fix Dashboard Container Not Mounting & Added extra info to user prop

William-Unosquare:
- Fixes an issue with the Dashboard's container, where it wasn't mounting because of the way the HOC's were composed.
- More user information is now available in the user prop in the Dashboard, use this to pass down to smaller components on the page.
- Adds lodash for better readability around HOC composition.


BrownMark (approved):
Looks grand, no issues here.

PR #18, opened 2018-05-25
Add Leave component

fraser-uno:
Bit of a restructure but as we discussed in the stand up today I think we were happy to copy ideas (steal) from the Reduction Dashboard example.

Added a Leave Component which will hold:
- Number of Holidays left
- Number of Holidays taken
Rather than having 2 separate components TakenLeave and BookedLeave.

It's not quite finished as the holidays that a user has requested have not yet been added. I think we should discuss how this will be implemented. (Haven't added any styling yet either)

Maybe plug the data from the Database into a React Calendar, so the user can switch month to month and have a visual view of holidays taken

### DIFF
--- a/web-app/src/components/Leave/container.js
+++ b/web-app/src/components/Leave/container.js
@@ -25,8 +25,6 @@ export default Wrapped =>
             return this.isDateInThePast(hol.date);
           });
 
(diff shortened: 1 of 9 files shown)


BrownMark (approved):
Looks good Fraser.

PR #20, opened 2018-05-29
S1.2.3 Request holiday feature built for single day requests

William-Unosquare:
- Adds RangeDayPicker component, which allows you to pick a range of dates, or a single date.
- Adds RequestHoliday component which is the UI and logic for requesting a holiday. The current set up should allow for the user to request a single-day holiday or a half-day holiday. (Pending back-end changes for multi-day holidays)

This is not in a modal yet, but I wanted to get the PR up so if anyone needs to create a holiday they can do it within the app.


web-app/src/components/RequestHoliday/container.js
```diff
@@ -1,169 +1,115 @@
...
+      );
 
-    buildHoliday(date) {
-      return {
-        date: date,
-        employee: this.props.user,
+      const request = {
+        date: formattedHolidayDate,
+        dateCreated: formattedTodayDate,
+        halfDay: this.state.halfDayChecked,
```
BrownMark, line 75: This should be changing very soon. But works for now and it's easily extended.

BrownMark (approved):

PR #21, opened 2018-05-30
Feature/add card styled component

fraser-uno:
Just added a Styled Component for the total leave displayed. 

Card layout similar to the Reduction Dashboard example we looked at


BrownMark (approved):
Looks good Fraser.

PR #22, opened 2018-05-30
add user details card to Dashboard

fraser-uno:
Added User Details card to the dashboard
Not sure what information we will display here so just added a couple of things for the time being

Had to change the prop name in the Dashboard from user -> userDetails
As user is already passed down from the withAuth component (which just contains the decoded JWT)


BrownMark (approved):
Looks good Fraser, no comments from me :)

PR #25, opened 2018-05-31
Feature/s1.2.5 cancel holiday request

peter-drysdale:
Cancel holiday calls in place (awaiting endpoints), confirmation checks added to holiday cancel and user sign out.


ghost:
Should we be ignoring the dist/bundle.js file?

BrownMark:
@girwinuno I think we have the dist folder in .gitignore, but it was done after it was checked in. we should remove it from the repo

web-app/src/components/BookedLeave/container.js
```diff
@@ -34,13 +38,71 @@ export default Wrapped =>
         });
     }
 
+    canCancelHoliday = (holiday) => {
```
bgribz, line 37: It'd be nicer to have this arrow function style for `cancelHoliday` and `handleCancelHoliday` as well - it would get rid of the icky binding in the constructor :)

web-app/src/components/BookedLeave/container.js
```diff
@@ -15,8 +15,12 @@ export default Wrapped =>
       this.state = {
         bookedHolidays: [],
       };
+
+      this.cancelHoliday = this.cancelHoliday.bind(this);
+      this.handleCancelHoliday = this.handleCancelHoliday.bind(this);
```
William-Unosquare: Remove the bind lines here. The reason these lines are necessary is to bind the `this` context to the component's class. Another (and better) way to do this with ES6 is to use the arrow syntax when creating these methods, which keeps the `this` context of the parent (in this case the component class). 

Eg:

```
cancelHoliday = holiday => {
 ...
}
```

web-app/src/components/BookedLeave/container.js
peter-drysdale: and that's done and pushed :-)

ghost:
Looking good Peter, looking good.

ghost (approved):
Nice!

web-app/src/components/BookedLeave/container.js
ghost: 👍

William-Unosquare (changes requested):
Just a quick ES6 change. Tried to explain as best as I could what's going on with this. Otherwise looks good to me. 👍 

Edit: 

Looks like I was beat to it, approved :)

William-Unosquare (approved):

ghost:
The rumours are @peter-drysdale  that you love JS and want to be a purely sole front end developer.....welcome to the community.

peter-drysdale:
I feel so warm and fuzzy inside!

PR #26, opened 2018-06-01
Feature/s1.1.1 create user

William-Unosquare:
Adds a component that contains a form which allows the creation of a new user. It has some error handling and success logic which should be good enough for MVP.

Because we don't know where this is going yet, it's not visible in the component but can be added wherever needed.

### DIFF
--- a/web-app/src/components/CreateUser/index.jsx
+++ b/web-app/src/components/CreateUser/index.jsx
@@ -107,7 +107,7 @@ const CreateUser = props => {
           <Button label="Create User" onClick={submitForm} disabled={loading} />
         </form>
         <Errorbox error={error} label="Error creating user" />
-        {success ? <p>User created successfully!</p> : null}
+        {success && <p>User created successfully!</p>}
       </Container>
     </Card>
   );
--- a/web-app/src/components/CreateUser/index.jsx
+++ b/web-app/src/components/CreateUser/index.jsx
@@ -1,0 +1,126 @@
+import React from 'react';
+import { PropTypes as PT } from 'prop-types';
+import { Container } from './styled';
+import container from './container';
+import { Card, Button, Errorbox } from '../common';
+import DatePicker from 'react-datepicker';
+
+const CreateUser = props => {
+  const {
+    formData,
+    submitForm,
+    formChanged,
+    startDateChanged,
+    error,
+    success,
+    loading,
+  } = props;
+  const {
+    forename,
+    surname,
+    email,
+    password,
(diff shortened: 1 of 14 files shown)


web-app/src/components/common/Button/index.js
```diff
@@ -2,9 +2,13 @@ import React from 'react';
 import { PropTypes as PT } from 'prop-types';
 import { Container } from './styled';
 
-const Button = ({label, onClick}) => (
-  <Container onClick={onClick}>{label}</Container>
-);
+const Button = props => {
+  return (
+    <Container onClick={props.onClick} {...props}>
```
bgribz: Can probs remove this `onClick` and leave it as `{...props}`

web-app/src/components/common/Button/styled.js
```diff
@@ -1,11 +1,20 @@
...
   text-align: center;
-  box-shadow: 0px 2px 2px rgba(0,0,0,0.1);
+  box-shadow: 0px 2px 2px rgba(0, 0, 0, 0.1);
   padding: 10px;
   border-radius: 4px;
+  width: 100%;
+  border: none;
+  font-size: 1em;
+  font-weight: bold;
+  ${({ disabled }) =>
```
bgribz: Could we integrate this into the `background` prop above?

web-app/src/components/common/Errorbox/index.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { PropTypes as PT } from 'prop-types';
+import { ErrorContainer } from './styled';
+
+const Errorbox = props => {
+  return props.error ? (
```
bgribz: Nicer to do an implicit return, and also drop the ternary for `props.error && ...`

web-app/src/components/CreateUser/index.jsx
```diff
@@ -0,0 +1,126 @@
...
+            <DatePicker
+              selected={startDate}
+              onChange={startDateChanged}
+              disabled={loading}
+            />
+          </label>
+          <Button label="Create User" onClick={submitForm} disabled={loading} />
+        </form>
+        <Errorbox error={error} label="Error creating user" />
+        {success ? <p>User created successfully!</p> : null}
```
bgribz: `success && ...`

bgribz (changes requested):
Few style changes but otherwise great.

bgribz (approved):

PR #28, opened 2018-06-04
Feature/show holidays on calendar and list

fraser-uno:
Added in the dates to the Big Calendar
Also added a Modal that lists all the holidays you've requested. 

Just showing the dates atm, not plugging in the status of the request (Requested, Pending etc) that we may use for styling events on the calendar


web-app/.env
```diff
@@ -1 +1 @@
-DOMAIN='http://localhost:8081'
\ No newline at end of file
+DOMAIN='http://localhost:8081'
```
bgribz: Is there something fishy going on with our `.gitignore`?

web-app/src/components/Leave/index.jsx
```diff
@@ -2,16 +2,36 @@ import React from 'react';
 import { PropTypes as PT } from 'prop-types';
 import container from './container';
 import { Card } from '../common/';
+import { Button } from '../common'
+import { Container, Content } from '../common/Modal/styled';
```
bgribz: It'd be better if we exported the Modal as its own component from `common`, rather than grabbing the styled components that it would use - if that makes sense.

web-app/src/pages/Dashboard/container.js
```diff
@@ -27,11 +28,15 @@ const DashboardContainer = Wrapped =>
     }
 
     componentDidMount() {
+      var self = this;
```
William-Unosquare: What's this doing?

web-app/src/pages/Dashboard/container.js
```diff
@@ -42,6 +47,39 @@ const DashboardContainer = Wrapped =>
         });
     }
 
+    getTakenHolidays() {
+      var self = this;
```
William-Unosquare: 👀 ?

web-app/src/pages/Dashboard/container.js
bgribz: I don't think this should be needed. Or, if it is, let's refactor so it's not lol

bgribz (changes requested):
Just a few things I noticed at a quick glance.

web-app/src/pages/Dashboard/container.js
```diff
@@ -42,6 +47,39 @@ const DashboardContainer = Wrapped =>
...
+          });
+        });
+    }
+
+    formatDates(events){
+        var eventsForCalendar = events.map(hol => {
+          return {
+           id: hol.holidayId,
+           title: `${hol.employee.forename} - Annual Leave`,
+           allDay: hol.halfDay ? false : true,
```
William-Unosquare: Use `allDay: !hol.halfDay`

William-Unosquare (changes requested):

web-app/src/pages/Dashboard/container.js
fraser-uno: Oh shit forgot to refactor this. 
I had a Promise in there which made 'this' refer to the Promise

fraser-uno: Yep, forgot to refactor that after removing a promise that wasn't needed

William-Unosquare (approved):

bgribz (approved):

PR #29, opened 2018-06-04
Fix/login 401 bug

William-Unosquare:
- Fixes issue where our Axios instance attempted to retrieve the token from the localStorage once, leading it to stay undefined if the user wasn't logged in when this was attempted. There's now an interceptor (think middleware for our requests) that re-retrieves the token.

- A bunch of code refactoring around how we got the user's info. Namely, it all comes from the initial dashboard `getUserInfo` call, instead of bits being hacked together from the email. The `getUserInfo` request now uses an ID call instead of getUserByFirstnameAndLastname.


web-app/src/components/common/Errorbox/styled.js
```diff
@@ -1,18 +1,25 @@
 import styled from 'styled-components';
 
 export const ErrorContainer = styled.div`
+<<<<<<< HEAD
```
bgribz: Missed a wee conflict here

web-app/src/hoc/withAuth.js
```diff
@@ -36,12 +32,7 @@ const withAuth = AuthComponent => {
         return <Redirect to="/login" />;
       }
       if (this.state.user) {
-        return (
-          <AuthComponent
-            user={{ ...this.state.user, ...UserInfo }}
-            {...this.props}
-          />
-        );
+        return <AuthComponent user={{ ...this.state.user }} {...this.props} />;
```
bgribz: Can this not just be `user={this.state.user}`? Or even `{...this.state}`

bgribz (changes requested):
Just a few things I noticed looking over this

web-app/src/components/common/Errorbox/styled.js
William-Unosquare: Good spot... Don't know why this hasn't been picked up by my editor or Github 😕

bgribz (approved):

PR #30, opened 2018-06-05
Fix/hidden conflicts

William-Unosquare:
A hotfix because more hidden conflicts that weren't flagged by anything popped up.


bgribz:
This seems to be adding conflicts :sweat_smile:

PR #31, opened 2018-06-05
Request/Edit/Cancel holiday

bgribz:
Adds functionality for
- Requesting single or multiple holidays
- Editing single holiday
- Cancelling single holiday

But it looks hella ugly, both on the UI and in the code :upside_down_face:

### DIFF
--- a/web-app/src/pages/Dashboard/container.js
+++ b/web-app/src/pages/Dashboard/container.js
@@ -149,9 +149,10 @@ const DashboardContainer = Wrapped =>
       const { booking, userDetails } = this.state;
       const request = [];
 
-      while (booking.start.diff(booking.end) <= 0) {
+      for (let i = 0; i <= booking.end.diff(booking.start, 'days'); i++) {
+        const start = booking.start.clone();
         request.push({
-          date: booking.start.format('YYYY-MM-DD'),
+          date: start.add(i, 'days').format('YYYY-MM-DD'),
           dateCreated: moment().format('YYYY-MM-DD'),
           halfDay: booking.isHalfday,
           holidayId: 0,
@@ -173,7 +174,6 @@ const DashboardContainer = Wrapped =>
             statusDescription: 'Inactive',
           },
         });
-        booking.start.add(1, 'days');
       }
 
       requestHoliday(request).then(() => {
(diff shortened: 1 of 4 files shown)


web-app/src/pages/Dashboard/container.js
```diff
@@ -67,10 +71,12 @@ const DashboardContainer = Wrapped =>
           allDay: !hol.halfDay,
           start: new moment(hol.date, 'YYYY-MM-DD'),
           end: new moment(hol.date, 'YYYY-MM-DD'),
+          isHalfday: hol.halfDay,
+          ...hol,
         };
       });
 
-      return eventsForCalendar;
+      return eventsForCalendar.filter(x => x.holidayStatusId !== 3);
```
BrownMark: Can we use a constant to check here

web-app/src/services/holidayService.js
```diff
@@ -2,18 +2,16 @@ import axios from '../utilities/AxiosInstance';
...
-export const requestHoliday = holiday => {
-  return axios.post('/holidays/', holiday);
-};
-
-export const requestHolidays = holidays => {
-  return axios.post('/holidays/createMultiple/', holidays);
+export const requestHoliday = holidays => {
+  return holidays.length === 1 ?
+    axios.post('/holidays/', holidays[0]) :
+    axios.post('/holidays/createMultiple/', holidays);
```
BrownMark, line 14: I don't think it makes sense for two calls for this, passing an array by default up to the server makes more sense. This shouldn't be a big change on both sides.

web-app/src/pages/Dashboard/container.js
```diff
@@ -139,13 +145,74 @@ const DashboardContainer = Wrapped =>
...
+        this.closeModal();
+      });
     };
 
+    updateHoliday = (cancel) => {
+      const { booking, userDetails } = this.state;
+      const request = {
+        date: booking.start.format('YYYY-MM-DD'),
+        dateCreated: moment().format('YYYY-MM-DD'),
+        halfDay: booking.isHalfday,
```
BrownMark, line 193: The new changes to the server side will massively reduce this request, but works for now

BrownMark (changes requested):
Just minor change for a holiday status constant and some notes

BrownMark (approved):

PR #32, opened 2018-06-05
Feature/admin panel

William-Unosquare:
chrome_2018-06-05_13-12-49

- Only shows up on nav bar for users with admin role.
- Users who navigate there directly will be redirected to the dashboard. (Or login, if not logged in)
- Contains the previously made CreateUser component which is fully functional.


BrownMark (approved):
Looks good William

PR #33, opened 2018-06-06
Admin Page - List Users & ability to change employee status

William-Unosquare:
chrome_2018-06-06_12-06-51

Adds a list of users to the admin page that we can use as a pathway into editing users, or viewing their holidays. The archive/activate button will allow you to set a user as inactive or active, but this status currently doesn't effect anything in-app. (Finishes S1.1.2)

If this stays we will need to work out some pagination down the line.


BrownMark (changes requested):
Looks great man, amazing amount of work in no time at all. just one comment basically.

web-app/src/components/UserListing/index.jsx
```diff
@@ -0,0 +1,47 @@
...
+      <tbody>
+        <tr>
+          <th>Status</th>
+          <th>Name</th>
+        </tr>
+        {users.map(user => (
+          <tr key={user.employeeId}>
+            <td>
+              <ActiveDot active={user.employeeStatusId === 1} />
+              {user.employeeStatusId === 1 ? 'Active' : 'Inactive'}
```
BrownMark: can we use a constant for the employee status instead of 1 (we use employeeStatus.ACTIVE) elsewhere

BrownMark (changes requested):

BrownMark (approved):
Looks good

PR #37, opened 2018-06-11
Remote working status/Team view

bgribz:
- Adds event styling for remote working status
- Adds team holiday view

The remote working status is hacked in at the minute (there's like a 30% chance the event will be a remote working one just for demoing). Waiting on the back end changes around adding a remote working table before making it any more functional. 

The team view is just a list of events grouped by employee ID, could probably do with some nice styling at some stage.


web-app/src/utilities/holidayStatus.js
```diff
@@ -2,4 +2,5 @@ export default {
   PENDING: 1,
   APPROVED: 2,
   REJECTED: 3,
+  WFH: 4,
```
William-Unosquare, line 5: I'd maybe change this ID to equal 10, in case we need to add more holiday statuses, and just to create a bit of seperation between a holiday and wfh event.

William-Unosquare (commented):
Looks good! Just one comment above

web-app/src/utilities/holidayStatus.js
bgribz: This is all going to change very soon anyway - I imagine it will be working off an `eventType` constant instead to accommodate working remotely vs a holiday. 

I think it's fine for now - 10 seems as arbitrary as 4 - this PR is just to show how remote working could be worked into/displayed on the calendar.

William-Unosquare: Cool, makes sense

William-Unosquare (approved):

web-app/src/components/TeamView/index.js
```diff
@@ -0,0 +1,77 @@
...
+import { PropTypes as PT } from 'prop-types';
+import moment from 'moment';
+import _ from 'lodash';
+
+class TeamView extends Component {
+  getEventsByEmployee = () => {
+    const events =
+      _.sortBy(this.props.events, (e) => e.start.toDate())
+        .filter(x =>
+          x.holidayStatusId !== 3
```
BrownMark: Let's keep the constants rather than magic numbers,

web-app/src/components/common/Event/index.js
```diff
@@ -1,10 +1,14 @@
 import React from 'react';
 import { PropTypes as PT } from 'prop-types';
+import FontAwesomeIcon from '@fortawesome/react-fontawesome';
+import { faHome, faPlane } from '@fortawesome/fontawesome-free-solid';
 import { Container } from './styled';
 
 const Event = ({children, event}) => (
   <Container status={event.holidayStatusId} onClick={children.props.onClick}>
-    {`${event.employee.forename} ${event.employee.surname}`}
+    {event.holidayStatusId === 4 && <FontAwesomeIcon icon={faHome}/>}
```
BrownMark: Same again with constants, even if it is potentially changing.

web-app/src/pages/Dashboard/container.js
```diff
@@ -72,7 +73,9 @@ const DashboardContainer = Wrapped =>
           start: new moment(hol.date, 'YYYY-MM-DD'),
           end: new moment(hol.date, 'YYYY-MM-DD'),
           isHalfday: hol.halfDay,
+          isWFH: false,
           ...hol,
+          holidayStatusId: Math.random() < 0.3 ? 4 : hol.holidayStatusId,
```
BrownMark, line 78: I'm not quite sure what this is about? Basically this has a 3/10 chance of being WFH?

BrownMark (changes requested):

web-app/src/pages/Dashboard/container.js
bgribz: Yep! Since the backend doesn't support remote working events yet I added this so a few events on the calendar would be remote. Again, just for demonstration.

BrownMark (approved):

PR #38, opened 2018-06-12
Feature/s1.5.4 add testsfor components

fraser-uno:
Some unit tests for some of the services, common components, containers and utilities

It's been a while since I've tested React components so if there's any bad practices which I'm sure there will be, then feel free to comment.

Also, in relation to the components wrapped in a container, what would be the best way to test these indepdendently?


web-app/__tests__/common/Button.js
```diff
@@ -0,0 +1,23 @@
...
+describe('Button', () => {
+  it('renders correctly', () => { 
+
+    const obj = { onButtonClick : () => {} };
+    const spy = jest.spyOn(obj, 'onButtonClick');
+
+    const wrapper = shallow(<Button id="button" label="Button" onClick={obj.onButtonClick}/>);    
+    
+    wrapper.find('#button').simulate('click');
+    expect(spy).toHaveBeenCalled();
```
William-Unosquare, line 21: I would split this into it's own it() call, as it has nothing to do with rendering correctly

web-app/__tests__/containers/CreateUser.js
```diff
@@ -0,0 +1,87 @@
...
+    const Container = createUserContainer(CreateUser);
+    const wrapper = shallow(<Container />);
+
+    userService.createUser = jest.fn(() => Promise.resolve({}));
+
+    await wrapper.instance().handleFormSubmit();
+    
+    expect(wrapper.state('loading')).toBe(false);
+    expect(wrapper.state('success')).toBe(true);
+    checkIfFormEqualsInitialState(wrapper);
```
William-Unosquare: Would it be worth splitting this up into 2 tests: "does form submit successfully (check the success state)" and "does form set state back to initial state on submission"? I think it should be more granular.

web-app/__mocks__/sweetalert2.js
```diff
@@ -0,0 +1 @@
+export default jest.fn(() => {});
```
bgribz: Is there any reason for moving these from `setupTests.js`?

web-app/__tests__/containers/Login.js
```diff
@@ -0,0 +1,39 @@
...
+    const Container = loginContainer(Login);
+    const wrapper = shallow(<Container history={ history }/>);
+
+    userService.userLogin = jest.fn(() => Promise.reject( { message: 'error' }));
+
+    wrapper.setState({ email: 'test@test.com', password: 'pass' });
+
+    await wrapper.instance().handleSubmit({ preventDefault: () => {} });
+    // expect(Swal).toHaveBeenCalled();
+  });
```
William-Unosquare: Think you've left an unfinished/broken test in

web-app/src/components/Leave/index.jsx
```diff
@@ -3,7 +3,7 @@ import { PropTypes as PT } from 'prop-types';
 import container from './container';
 import { Card, Button, Modal } from '../common';
 
-export const Leave = props => {
+const Leave = props => {
```
bgribz: Exporting the component here would allow us to test the component itself without the wrapping container. In your test file you'd just have to do a named import e.g. `import { Leave } from 'xyz`

bgribz (changes requested):
Looks good, just a couple of queries.

web-app/__mocks__/sweetalert2.js
fraser-uno: I was experimenting with mocks... Seems to be best practice to move all mocks into a `__mocks__` folder

web-app/src/components/Leave/index.jsx
fraser-uno: Yeah this is what I was wondering about. But in our index.js file where we export all the components we do: 
 `export { default as Leave } from './Leave';`

Would this be conflicting when importing?

web-app/setupTests.js
```diff
@@ -1,6 +1,6 @@
+require('dotenv').config();
 import Enzyme from 'enzyme';
 import EnzymeAdapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new EnzymeAdapter() });
-
-jest.mock('sweetalert2', () => 'Swal');
+// jest.mock('sweetalert2', jest.fn() );
```
William-Unosquare: Remove unused code

William-Unosquare (changes requested):
Few quick changes but awesome work man - it was really useful to look through this to see how it's done 👍

web-app/src/components/Leave/index.jsx
William-Unosquare: @fraser-uno Nope, because it's the default that's being retrieved, and `Leave` is just what you're assigning it to.

fraser-uno: Oh right okay, cool. We can do this with all components then, whole lot easier

bgribz: @fraser-uno - you're right, sorry, should have been clearer. When you're importing it in the test file, import it from the `Leave` directory rather than the `components` one.

So `import { Leave } from 'components/Leave'`. That'll import the export named `Leave` from the `index.js` file in `components/Leave`

Just for reference, 
`import Leave from 'components/Leave'` would grab the default export, which in this case would be the wrapped component.

Our big file of exports is just so that we can import multiple components in one line. I'm not very good at explaining things so if you want me to run through it with you give me a shout.

web-app/__tests__/containers/Login.js
fraser-uno: Yeah that mock wasn't behaving for me. Going to have to play about with it and spyOn it properly

web-app/__tests__/containers/CreateUser.js
fraser-uno: Yeah, can do. Makes sense

web-app/src/components/Leave/index.jsx
fraser-uno: @bgribz no no that's a good explanation. First time I've come across project with default and named exports so it was bit confusing to get my head around!

William-Unosquare (approved):

BrownMark (approved):

bgribz (approved):

PR #41, opened 2018-06-13
Feature/improved team view

William-Unosquare:
chrome_2018-06-13_12-14-39

Updates team view so that each employee has their own holiday table similar to those found in the admin panel. As a wee bonus I've also added Open Sans as the default font.


web-app/src/components/TeamView/index.js
```diff
@@ -3,67 +3,97 @@ import { PropTypes as PT } from 'prop-types';
...
       return acc;
     }, {});
 
     return eventsByEmployee;
-  }
+  };
+
+  statusToText = status => {
+    switch (status) {
+      case 1:
```
BrownMark, line 35: Can we use the constants here?

BrownMark: Might also be better to separate this out, i suspect it will be reused

BrownMark (changes requested):
Just some minor changes, nothing too taxing :)

BrownMark (approved):

PR #44, opened 2018-06-14
Feature/event icons

William-Unosquare:
- Adds icons to the legend, and all calendar events.
- Changes the legend test as there were a few checks in there that didn't make sense to do as later tests would fail anyway if the requirements weren't met.
- Changed "WFH" to "Working Remotely".


web-app/src/components/common/Errorbox/styled.js
```diff
@@ -1,18 +1,25 @@
 import styled from 'styled-components';
 
 export const ErrorContainer = styled.div`
+<<<<<<< HEAD
```
bgribz: :sob:

web-app/src/utilities/holidayStatus.js
```diff
@@ -1,8 +1,31 @@
+import React from 'react';
+import FontAwesomeIcon from '@fortawesome/react-fontawesome';
+import {
+  faThumbsUp,
```
bgribz, line 4: I'm not sure everything needs an icon - that's what the colours are for. Remote working will need one because it can be one of two types, but I think the colour scheme is clear enough.

Thoughts?

bgribz (dismissed):

web-app/src/utilities/holidayStatus.js
William-Unosquare: I think because remote working will need one that we should keep it consistent & always supply an icon. Also because I wanna power up my reply to you I'm gonna throw some accessibility talk in here and add that colours aren't useful if you're colour-blind.

BrownMark (approved):
Looks good Will.

PR #49, opened 2018-06-19
Feature/user page

William-Unosquare:
- Adds a new page `/user/:userId` to the app that renders details about a specific user. Right now this is an admin only page. It would be good to expand it so that a team leader could view their team members on this page.

chrome_2018-06-19_13-31-58

- I got fed up with making similar holiday tables repeatedly soooo I gave that job to a `<HolidayList />` component. Basically, you feed it an array of holidays for it to use. Then you define the columns and actions with 2 other array props. This is how it looks for the table in the screenshot above:

```
<HolidayList
    holidays={profileHolidays}
    columns={['status', 'start', 'end', 'requested']}
    actions={['approve', 'reject']}
/>
```

- I added an interceptor that gets triggered when a list of holidays are called, that adds a `start` `end` and `requested` field to the data. These are basically moment.js objects for the holiday dates, making our lives easier when it comes to formatting dates.

### DIFF
--- a/web-app/src/components/common/HolidayList/index.jsx
+++ b/web-app/src/components/common/HolidayList/index.jsx
@@ -12,13 +12,14 @@ const buildTableHeaders = (columns) => {
 
 const buildTableValues = (columns, holiday) => {
   const { forename, surname } = holiday.employee;
-  const { holidayId, date, dateCreated, holidayStatusId } = holiday;
+  const { holidayId, start, end, requested, holidayStatusId } = holiday;
 
   const values = {
     status: <td key={`${holidayId}status`}><StatusDot status={holidayStatusId} />{statusText[holidayStatusId]}</td>,
     employee: <td key={`${holidayId}employee`}>{`${forename} ${surname}`}</td>,
-    date: <td key={`${holidayId}date}`}>{`${date[2]}/${date[1]}/${date[0]}`}</td>,
-    created: <td key={`${holidayId}created`}>{`${dateCreated[2]}/${dateCreated[1]}/${dateCreated[0]}`}</td>,
+    start: <td key={`${holidayId}start}`}>{start.format('Do MMM YYYY')}</td>,
+    end: <td key={`${holidayId}end}`}>{end.format('Do MMM YYYY')}</td>,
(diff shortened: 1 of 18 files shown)


fraser-uno (commented):
Really like this idea, from an Admin perspective this would be quite useful

web-app/src/components/common/HolidayList/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+  }, []);
+};
+
+const buildTableActions = (props, holiday) => {
+  const {actions, approveHoliday, rejectHoliday } = props;
+  const { holidayId } = holiday;
+
+  const values = {
+    approve: <button key={`${holidayId}approve`} onClick={() => approveHoliday(holiday)}>Approve</button>,
+    reject: <button key={`${holidayId}reject`} onClick={() => rejectHoliday(holiday)}>Reject</button>,
```
fraser-uno, line 36: Great job on this. I think maybe because the direction this file is going is reusability, maybe these constants could be saved in a separate file?

web-app/src/components/common/HolidayList/index.jsx
```diff
@@ -0,0 +1,78 @@
...
+const buildTableValues = (columns, holiday) => {
+  const { forename, surname } = holiday.employee;
+  const { holidayId, date, dateCreated, holidayStatusId } = holiday;
+
+  const values = {
+    status: <td key={`${holidayId}status`}><StatusDot status={holidayStatusId} />{statusText[holidayStatusId]}</td>,
+    employee: <td key={`${holidayId}employee`}>{`${forename} ${surname}`}</td>,
+    date: <td key={`${holidayId}date}`}>{`${date[2]}/${date[1]}/${date[0]}`}</td>,
+    created: <td key={`${holidayId}created`}>{`${dateCreated[2]}/${dateCreated[1]}/${dateCreated[0]}`}</td>,
+  };
```
fraser-uno: Same with these?

fraser-uno (approved):

web-app/src/components/common/HolidayList/index.jsx
William-Unosquare: Yep. Barry also mentioned this. I'm going to rework it a little more.

BrownMark (approved):

PR #53, opened 2018-06-25
UI Refinements & React-table implementation for HolidayList

William-Unosquare:
- Refines the booking modal component.
- Refines the user info component.
- `react-table` is now being used in the HolidayList component.
  - Added custom sorting
  - Added custom filtering
- Replaced any table of holidays with HolidayList
- Updates tests


ghost:
I'll have a look through in a few blocks between doing project work. Going forward it might be easier for others + possibly for yourself if we could break the PR's down into smaller ones. Just so they can be reviewed + reverted (if needed) nicer. Possibly one PR for updating the styling for the booking modal, another for the work around the holiday list component etc.

web-app/src/pages/User/styled.js
```diff
@@ -22,7 +22,7 @@ export const Container = styled.div`
     margin-bottom: 5px;
   }
   h2 {
-    margin: 0;
+    margin: 0 0 1rem 0;
```
ghost: Some are .rem and some are .px is there a reason for that?

William-Unosquare: No reason! Will change.

web-app/src/components/HolidayList/TableValues.js
```diff
@@ -1,49 +1,52 @@
...
+  id: 'requestedDate',
+  Header: 'Requested Date',
+  'accessor': holiday => holiday.requested.format('D MMM YYYY'),
 };
 
-export default generateValue;
\ No newline at end of file
+export default {
+  status, startDate, endDate, requestedDate, employee,
+};
```
ghost: Small thing but if you could add a empty line to the bottom of the files then it'll help for linting.

web-app/src/components/UserDetails/index.jsx
```diff
@@ -26,9 +28,9 @@ export const UserDetails = props => {
   return (
     <Card>
       <StyleContainer>
-        <strong>Logged in as</strong>
-        <p>Name: {`${surname}, ${forename}`}</p>
-        <p>Role: {roleIdToText(employeeRoleId)} </p>
+        <strong>Welcome,</strong>
+        <strong>{`${forename} ${surname}`}</strong>
```
ghost, line 32: Such a small annoying thing I know, but anyway we could remove strong tags and use CSS. A lot nicer if we want to change it one go etc. If it can't be done though, fair enough.

ghost:
William that looks awesome man, not really sure about all of the business logic as I'm not on the project but seems like you've smashed it. Only thing I'd say was the line endings just need to be taken care of for when the CI stuff comes in and we add linting. Apart from that nice 👍

web-app/src/components/UserDetails/index.jsx
William-Unosquare: Ah, never even thought of that. I've always thought it was better semantically because of screen readers and such, but we don't need to worry about those here. 

It can be done and I've taken note for the future, but this component is going to be gone soon in favour including it somewhere in the new sidebar so I'll not worry about it in this case. 👍

BrownMark (approved):
Looks good William

William-Unosquare:
Added new lines & with that, I'm gonna merge.

PR #54, opened 2018-06-26
S1.6.2 - View amount of leave I can take.

kirsty-unosquare:
Displays the amount of leave the user can take ( displayed as remaining holidays ). Also configured the service call within home container.

### DIFF
--- a/NativeApp/src/components/User/container.js
+++ b/NativeApp/src/components/User/container.js
@@ -1,5 +1,5 @@
 import React, { Component } from 'react';
-import { getUserHolidays } from '../../services/userService';
+import { getTakenHolidays, getRemainingHolidays } from '../../utilities/holidays';
 
 export default Container => class extends Component {
   constructor(props) {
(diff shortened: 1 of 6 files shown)


NativeApp/src/components/User/container.js
```diff
@@ -1,27 +1,30 @@
...
     };
   }
 
   componentDidMount() {
-    getUserHolidays()
+    getTakenHolidays()
       .then(data => this.setState({ takenHolidays: data }));
+
+    getRemainingHolidays()
+      .then(data => this.setState({ remainingHolidays: data }));
```
d1wilko: generally a good practice to stay away from ambiguously named variables like `data`, I appreciate you've copied the code from above but this might be nicer as something like...
`.then(remainingHolidays => this.setState({ remainingHolidays }));`
could even change the code above 👀

d1wilko (approved):
Some really nice code - just one suggestion which I'll leave up to you :)

NativeApp/src/components/User/container.js
kirsty-unosquare: Totally agree with you. Also makes more sense and easier to understand what data your actually passing in.

PR #56, opened 2018-06-28
Feature/adjustments accomodating new nav

William-Unosquare:
- Changes various aspects of the UI to fit in with the new sidebar, but more work needs to be done.
  - The top of the calendar has been changed to remove the view swapping buttons, and to look just a bit nicer overall. Because of this, you can no longer get on to the personal and team pages as they're going to be moved to their own pages.
  - The sidebar on the dashboard has been removed. The user details component that was here is removed in favour of a profile page, and the legend is now at the bottom of the calendar.


BrownMark (approved):
Looks good :)

PR #59, opened 2018-06-29
changed to navbar to be stateful component

paulmatchett:


web-app/src/hoc/Layout/header.jsx
```diff
@@ -36,98 +32,75 @@ const Header = ({ history, isAuthenticated }) => {
     });
   };
 
-  const inAdminSection = (match, location) => {
-    const { pathname } = location;
-    if (pathname.indexOf('admin') != -1) {
-      return true;
+  const creatTopLevelNavLinks = headerLink => {
```
William-Unosquare: You've missed an 'e' in the const name here, I would fix this up in case someone trips on it later.

web-app/src/utilities/menuIcons.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
William-Unosquare: I think I get _why_ you've made this file, but it seems like an abstraction that we won't benefit much from because there's only ever going to be 1 component needing this data. I'd just use the <FontAwesomeIcon /> component directly in the menu.

web-app/src/hoc/Layout/container.js
```diff
@@ -0,0 +1,63 @@
...
+  faUnlockAlt,
+} from '@fortawesome/fontawesome-free-solid';
+
+export default Wrapped =>
+  class extends Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+        menuItems: [
+          {
```
William-Unosquare, line 14: This would be better placed in a `navConfig.js` file or something. Use state for stuff that you anticipate to change while the user is using the app. I don't think this will be changing.

Because of this, the whole container isn't needed.

William-Unosquare (changes requested):
This is great work man. It'll make changing the navigation far easier. Just a couple of changes which shouldn't take long.

web-app/src/hoc/Layout/container.js
William-Unosquare: Yep so just ignore this feedback - I can see why you've done this to support changes in the future, nice work man

BrownMark (approved):

William-Unosquare (approved):

PR #63, opened 2018-07-03
New Auth wrapper that retrieves users profile & stores it in redux

William-Unosquare:
- Gets rid of the old withAuth component which was outdated, and kinda a pain.
- Adds new auth component that will:
  - Authenticate the user / redirect them to the login if failure
  - Grab the user's profile, and store this in the redux store

This change means we don't need to wrap every page with an authentication hoc as it's done higher level and it means we can start using the redux store for grabbing the user's info.


BrownMark (approved):
I like it. Looks good to me

PR #65, opened 2018-07-03
Persist whether or not the nav drawer is open between page reloads

William-Unosquare:
- The drawer now uses localStorage to persist if it's open or not between page reloads.

### DIFF
--- a/web-app/src/hoc/Layout/container.js
+++ b/web-app/src/hoc/Layout/container.js
@@ -10,7 +10,7 @@ export default Wrapped =>
     constructor(props) {
       super(props);
       this.state = {
-        drawerOpen: localStorage.getItem('navDrawerOpen') == 'true' || false,
+        drawerOpen: localStorage.getItem('navDrawerOpen') == 'true',
         menuItems: [
           {
             name: 'Profile',
@@ -59,8 +59,9 @@ export default Wrapped =>
     }
 
     toggleDrawer = () => {
-      this.setState({ drawerOpen: !this.state.drawerOpen });
-      localStorage.setItem('navDrawerOpen', !this.state.drawerOpen);
+      this.setState({ drawerOpen: !this.state.drawerOpen }, () => {
+        localStorage.setItem('navDrawerOpen', this.state.drawerOpen);
+      });
     };
 
     render() {
--- a/web-app/src/hoc/Layout/container.js
+++ b/web-app/src/hoc/Layout/container.js
@@ -10,6 +10,7 @@ export default Wrapped =>
     constructor(props) {
       super(props);
       this.state = {
+        drawerOpen: localStorage.getItem('navDrawerOpen') == 'true' || false,
         menuItems: [
           {
             name: 'Profile',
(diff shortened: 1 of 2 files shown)


web-app/src/hoc/Layout/index.js
```diff
@@ -36,5 +42,7 @@ Layout.propTypes = {
   history: PT.object,
   isAuthenticated: PT.bool,
   children: PT.node,
+  drawerOpen: PT.bool,
+  toggleDrawer: PT.func,
```
d1wilko: Can we use isRequired or defaultProps going forward?

web-app/src/hoc/Layout/container.js
```diff
@@ -10,6 +10,7 @@ export default Wrapped =>
     constructor(props) {
       super(props);
       this.state = {
+        drawerOpen: localStorage.getItem('navDrawerOpen') == 'true' || false,
```
d1wilko: why `'true'` instead of `true` here?

I appreciate there may be a good reason for this, but it suggests a code smell to me at first glance

d1wilko (changes requested):
Looks good WIlliam, just two points :)

web-app/src/hoc/Layout/container.js
William-Unosquare: localStorage.getItem seems to return true as a string instead of a bool. It threw me off quite a bit too.

bgribz: I don't think there's a need for the `|| false` here

web-app/src/hoc/Layout/container.js
```diff
@@ -57,7 +58,18 @@ export default Wrapped =>
       };
     }
 
+    toggleDrawer = () => {
+      this.setState({ drawerOpen: !this.state.drawerOpen });
+      localStorage.setItem('navDrawerOpen', !this.state.drawerOpen);
```
bgribz: I think
```
this.setState(
    {drawerOpen: !this.state.drawerOpen}, () => 
    localStorage.setItem('navDrawerOpen', this.state.drawerOpen)
)
```
is safer here. Not sure we can guarantee that the state will (won't?) have been updated when setting the localStorage token.

bgribz (changes requested):

web-app/src/hoc/Layout/index.js
```diff
@@ -6,10 +6,16 @@ import { LayoutContainer, Input } from './styled';
...
+
   if (props.isAuthenticated) {
     drawer = (
       <Fragment>
-        <Input type="checkbox" id="toggle-drawer" />
+        <Input
+          type="checkbox"
+          id="toggle-drawer"
+          onChange={props.toggleDrawer}
+          checked={props.drawerOpen}
```
ghost: Probably good to destructure these props at the top plus readability for consistency across the app.

web-app/src/hoc/Layout/container.js
bgribz: Also probs best to use the strict comparison `===`

William-Unosquare: @bgribz If localStorage can't find what it's looking for it returns null. I'd rather keep the false fallback to avoid any issues cropping up from accessing null.

William-Unosquare: Would we agree on 
`drawerOpen: localStorage.getItem('navDrawerOpen') === 'true' ? true : false,`

That way there's no string being stored, and there's no null being stored in `drawerOpen`.

bgribz: But you wouldn't be setting `null`, you'd be setting the result of `null === 'true' => false`

William-Unosquare: gotcha - my head isn't in the game this morning 😄

d1wilko (approved):

bgribz (approved):

BrownMark (approved):

PR #67, opened 2018-07-04
Feat/s1.6.3 request for leave

kirsty-unosquare:
Requesting holidays is now working. The date picker looks horrible but gonna separate it into another branch, a lot of work still to do on it. But in this PR:
 - new modal wrapper within common
 - Booking modal component, shows after selecting a day on calendar.
 - New datepicker currently working only for ios devices. Will be continuing on another branch.
 - Logic for the requesting holidays.


NativeApp/src/components/Home/container.js
```diff
@@ -18,20 +21,112 @@ export default Container => class extends Component {
...
+        booking: {
+          ...prevState.booking,
+          endDate: formatEndDate,
+        },
+      }));
+    }
+
+    closeModal = () => {
+      const { showModal } = this.state;
+      this.setState({ showModal: !showModal });
```
d1wilko: should we change `!showModal` to `false` here?
I appreciate that if the modal is open `showModal` should be `true` - but I'm just wary incase future code messes that state setting up.

NativeApp/src/components/Home/container.js
```diff
@@ -18,20 +21,112 @@ export default Container => class extends Component {
...
+      dates.push(endDate);
+
+      return dates;
     }
 
     formatDate = data => data.reduce((obj, item) => {
       const holidayStatus = this.holidayStatus(item.holidayStatusId);
-      obj[item.start] = { textColor: 'white', color: holidayStatus };
+      const dates = this.enumerateDaysBetweenDates(item.start, item.end);
+      for (let i = 0; i < dates.length; i += 1) {
```
d1wilko: could this not be done as a foreach loop? much nicer syntax

d1wilko (changes requested):
Nice code - just 2 suggestions

NativeApp/src/components/Home/container.js
kirsty-unosquare: Good point! Safer to set it to false.

NativeApp/src/components/Home/container.js
kirsty-unosquare: Can't see why it couldn't.

kirsty-unosquare:
@d1wilko Got those changes up.

d1wilko (approved):
looks great :)

PR #68, opened 2018-07-04
Updates endpoints to work with new back-end and fixes any bugs raised from the update

William-Unosquare:
- Updates endpoint calls to send data matching the updated dtos.
- Adds a workaround currently for there being no employee object being turned from a holiday request.
- Fixes up multi-day holidays in the calendar... they now show as one block, or event, spanning multiple days.


web-app/src/hoc/Layout/container.js
```diff
@@ -0,0 +1,63 @@
...
+  faTachometerAlt,
+  faUnlockAlt,
+} from '@fortawesome/fontawesome-free-solid';
+
+export default Wrapped =>
+  class extends Component {
+    constructor(props) {
+      super(props);
+      this.state = {
+        menuItems: [
```
bgribz: Why is this being stored in state?

bgribz (changes requested):
Just a small query

web-app/src/hoc/Layout/container.js
William-Unosquare: I brought this up before with @paul-matchett in a previous PR. I suggested moving it into a nav config. His reasoning was because thinks it would be useful in the future if the user could modify their sidebar to suit their needs. (Pin pages, and whatnot)

I said it was grand because of the reasoning but looking back now I think it's maybe a bit premature to have it in the state.

bgribz: Yeah, I can't see it needing to be modified anytime soon, think it's best stored elsewhere 👍

William-Unosquare: @bgribz I'll fix this in another branch as it's not the focus of this PR.

bgribz (approved):

PR #70, opened 2018-07-05
Added controller for retrieving employee roles

peter-drysdale:
Added controller for retreiving employee roles, still using enums at this stage but available to front end devs for now


BrownMark:
I'm not sure what driver is for this.

peter-drysdale:
It wasn't there. If it shouldn't be there anyway that's fine, but if it ought to be there then there's a branch for it.

PR #71, opened 2018-07-05
Feature/input validation

paulmatchett:
Input validation added to booking modal


web-app/src/components/BookingModal/index.js
```diff
@@ -1,102 +1,131 @@
...
+    );
+
+  // const getTotalDays = () => {
+  //   if (booking.form.isWFH.value) {
+  //     return '0';
+  //   }
+  //   return (
+  //     Math.floor(booking.duration) / (booking.form.isHalfday.value ? 2 : 1)
+  //   );
+  // };
```
William-Unosquare: Remove unused code

web-app/src/components/BookingModal/index.js
```diff
@@ -1,102 +1,131 @@
...
+  showModal: PT.bool.isRequired,
+  booking: PT.object.isRequired,
+  closeModal: PT.func.isRequired,
+  formChanged: PT.func.isRequired,
+  requestHoliday: PT.func.isRequired,
+  cancelHoliday: PT.func.isRequired,
+  //changeEnd: PT.func,
+  //changeHalfday: PT.func,
+  //takenHolidays: PT.array,
+  //changeWFH: PT.func,
```
William-Unosquare: Remove unused code

William-Unosquare (changes requested):
Some commenting needing removed here. I'd hold off on merging this until we can go over this a bit more tomorrow. It's tough to put together in my head looking at these changes. But I strongly suspect a lot of this can be abstracted away into their own components in a more reusable way.

web-app/src/components/BookingModal/index.js
paulmatchett: I kept this commented as a why to remind me to ask you what it was for. The value that this function returns is already being handled in and contained booking.duration? But yes, we can talk about this tomorrow.

web-app/src/components/BookingModal/index.js
```diff
@@ -1,102 +1,118 @@
...
   } = props;
 
-  const getActions = () => (
-    booking.title ?
+  const getActions = () =>
+    booking.title ? (
       <ButtonWrap>
-        <Button id="updateHolidayBtn" onClick={updateHoliday} label={'Update'} />
-        <Button id="cancelHolidayBtn" onClick={cancelHoliday} label={'Cancel Booking'} />
+        <Button
```
ghost, line 30: Quite a big ternary here. This could be broken down a bit more with logic inside the button component.

paulmatchett: Yeah, I totally agree. I'll update it now. thanks

web-app/src/components/BookingModal/index.js
```diff
@@ -1,102 +1,118 @@
...
+    let cssClass = '';
+    if (booking.duration == 0) {
+      cssClass = 'workingFromHome';
+    } else if (booking.duration == 0.5) {
+      cssClass = 'bookingHalfDay';
+    }
+    return cssClass;
   };
 
+  let form = (
```
bgribz, line 62: Could use `const` here as it doesn't seem to ever change. But better again, this should be split out into its own component.

web-app/src/pages/Dashboard/container.js
```diff
@@ -219,25 +426,23 @@ const DashboardContainer = Wrapped =>
...
+            formChanged={(event, id) => this.handleFormChange(event, id)}
             userDetails={this.state.userDetails}
             requestHoliday={this.submitHolidayRequest}
             takenHolidays={this.state.takenHolidays}
             updateHoliday={() => this.updateHoliday(false)}
             cancelHoliday={() => this.updateHoliday(true)}
-            changeWFH={this.changeWFH}
             {...this.state}
             {...this.props}
           />
```
bgribz, line 448: This isn't really a concern for this PR, but holy crap! A ~450 line JS file!

I think this is an indication that we need to review how we're composing components.

bgribz (changes requested):

web-app/src/pages/Dashboard/container.js
paulmatchett: Totally agree. I'm new to react so any guidance would be greatly appreciated. ps this was a large file before I came along.

web-app/src/components/BookingModal/index.js
paulmatchett: Will update this. thanks

PR #72, opened 2018-07-05
Feat/configure booking nav

kirsty-unosquare:
Kinda went down a rabbit hole with this one. Originally was meant to set the date picker up nicely but then came across an issue. Currently the requesting holidays was built using a modal, the date picker will also be a modal. A modal on top of a modal will only cause problems in the future, plus is genuinely   bad practise.

So in the PR:
- I've separated the booking component into it's own screen.
- Update the nav config for new home stack navigator. Which includes the Home and Booking screen.
- UI updates to Tab bar.
- New `didFocus` listener added to home container. Updates the calendar screen every time it's focused.

### DIFF
--- a/NativeApp/src/components/Booking/container.js
+++ b/NativeApp/src/components/Booking/container.js
@@ -1,0 +1,101 @@
+import React, { Component } from 'react';
+import { Alert } from 'react-native';
+import { PropTypes as PT } from 'prop-types';
+import moment from 'moment';
+import { userProfile } from '../../utilities/currentUser';
+import { requestHolidays } from '../../services/holidayService';
+
+
+export default Container => class extends Component {
+  static propTypes = {
+    navigation: PT.shape({
+      navigate: PT.fun,
+    }).isRequired,
+  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      booking: {
+        startDate: '',
+        endDate: '',
+        halfDay: false,
+      },
+      user: {},
+    };
+  }
+
+  componentDidMount() {
+    const { navigation } = this.props;
+    const chosenDate = navigation.getParam('date', '');
+
+    userProfile()
+      .then(user => this.setState({ user }));
+
+    this.setState({
+      booking: {
+        startDate: chosenDate,
+        endDate: chosenDate,
+      },
+    });
+  }
+
+  changeStartDate = (date) => {
+    const formatDate = moment(date).format('YYYY-MM-DD');
+    this.setState(prevState => ({
+      booking: {
+        ...prevState.booking,
+        startDate: formatDate,
+      },
+    }));
+  }
+
+  changeEndDate = (endDate) => {
+    const formatEndDate = moment(endDate).format('YYYY-MM-DD');
+    this.setState(prevState => ({
+      booking: {
+        ...prevState.booking,
+        endDate: formatEndDate,
+      },
+    }));
+  }
+
+  submitRequest = () => {
+    const { booking, user } = this.state;
+    const { navigation } = this.props;
+
+    const request = {
+      dates: [
+        {
+          endDate: booking.endDate,
+          halfDay: false,
+          startDate: booking.startDate,
+        },
+      ],
+      employeeId: user.employeeId,
+    };
+
+    requestHolidays(request)
+      .then(() => {
+        navigation.pop();
+      })
+      .catch(e => Alert.alert(
+        'Could not request holidays',
+        e.message,
+      ));
+  }
+
+  render() {
+    const { booking } = this.state;
+
+    return (
+      <Container
+        startDate={booking.startDate}
+        endDate={booking.endDate}
+        submitRequest={this.submitRequest}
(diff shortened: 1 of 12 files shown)


NativeApp/src/screens/Booking.js
```diff
@@ -0,0 +1,23 @@
...
+  static propTypes = {
+    navigation: PT.shape({
+      navigate: PT.func,
+    }).isRequired,
+  }
+
+  render() {
+    const { navigation } = this.props;
+
+    return (
```
d1wilko: I might be wrong here - but can this be written as this? `return <Booking navigation={navigation} />`

It would just be a bit cleaner

d1wilko (approved):
Looks good, 1 suggestion but approving anyway

NativeApp/src/screens/Booking.js
kirsty-unosquare: Your totally right. Thought the linter would complain but didn't. Gonna convert the other screens as well.

NativeApp/src/components/Booking/container.js
```diff
@@ -0,0 +1,101 @@
...
+import { PropTypes as PT } from 'prop-types';
+import moment from 'moment';
+import { userProfile } from '../../utilities/currentUser';
+import { requestHolidays } from '../../services/holidayService';
+
+
+export default Container => class extends Component {
+  static propTypes = {
+    navigation: PT.shape({
+      navigate: PT.fun,
```
bgribz, line 12: Not sure you need to go this deep with propTypes. Especially since the navigation property is coming from a third party library and isn't something we can control.

But if you want to keep this, I think this should be `navigate: PT.func`?

NativeApp/src/components/Booking/container.js
```diff
@@ -0,0 +1,101 @@
...
+      },
+      user: {},
+    };
+  }
+
+  componentDidMount() {
+    const { navigation } = this.props;
+    const chosenDate = navigation.getParam('date', '');
+
+    userProfile()
```
bgribz, line 32: Should this be renamed to `getUserProfile`? I think it would make more sense.

NativeApp/src/components/Booking/container.js
```diff
@@ -0,0 +1,101 @@
...
+      booking: {
+        startDate: chosenDate,
+        endDate: chosenDate,
+      },
+    });
+  }
+
+  changeStartDate = (date) => {
+    const formatDate = moment(date).format('YYYY-MM-DD');
+    this.setState(prevState => ({
```
bgribz, line 45: Is there any particular benefit to using `prevState` rather than `this.state`? Just a query as I've never seen `setState` used this way before!

NativeApp/src/components/Home/container.js
```diff
@@ -32,70 +30,34 @@ export default Container => class extends Component {
...
-        .then(data => this.setState({ takenHolidays: this.formatDate(data) }));
+      const { navigation } = this.props;
+
+      this.sub = navigation.addListener('didFocus', () => {
+        getTakenHolidays()
+          .then(data => this.setState({ takenHolidays: this.formatDate(data) }));
+
+        userProfile()
+          .then(user => this.setState({ user }));
+      });
```
bgribz: Is there any way to refactor this so that `setState` is only being called once?

bgribz (changes requested):
Few wee changes, but otherwise looks great.

d1wilko:
looks good, I'd say @bgribz should have a look see

NativeApp/src/components/Booking/container.js
kirsty-unosquare: Linter was complaining about this, forbids vague prop types. I can set the linter to ignore this but think I prefer keeping it in. What do you think?

NativeApp/src/components/Booking/container.js
kirsty-unosquare: This had to do with changing the date picker. I have two, start and end date pickers. Therefore each receive their own state for chosen date. I have to update the state when the date picker is changed. Problem was either the start date or end date state was disappearing on change.

The `prevstate` basically allows me to bring in prev state and I can update either end date or start date without erasing them.

NativeApp/src/components/Booking/container.js
kirsty-unosquare: I've actually got a service call called `getUserProfile`.  @bgribz can't think of another name for it. Any other options?

NativeApp/src/components/Booking/container.js
bgribz: I think we shouldn't be using AirBnB's linting rules for a start 🙃 

But that's just me, keeping it in and fixing the `PT.func` is grand!

NativeApp/src/components/Booking/container.js
bgribz: Oooh ok, em it's not terribly important, the code after makes it clear enough I guess!

bgribz (approved):

PR #75, opened 2018-07-10
Feat/datepicker config (ios)

kirsty-unosquare:
Configuration for ios date picker.
- Date picker now shows in a pop up modal at bottom of screen.
- Dates now displayed in the `customDatePickerIOS`

### DIFF
--- a/NativeApp/src/components/Common/CustomDatePickerIOS/index.js
+++ b/NativeApp/src/components/Common/CustomDatePickerIOS/index.js
@@ -66,16 +66,14 @@ class CustomDatePickerIOS extends Component {
             >
               <View style={styles.dateContainer}>
                 <View style={styles.background}>
-                  <View>
-                    <DatePickerIOS
-                      date={formatDate(chosenDate)}
-                      onDateChange={setDate}
-                      mode="date"
-                      minimumDate={formatDate(minimumDate)}
-                    />
-                    <View style={styles.buttonContainer}>
-                      <Button onPress={() => this.setModalVisible(false)} title="Done" />
-                    </View>
+                  <DatePickerIOS
+                    date={formatDate(chosenDate)}
+                    onDateChange={setDate}
+                    mode="date"
+                    minimumDate={formatDate(minimumDate)}
+                  />
+                  <View style={styles.buttonContainer}>
+                    <Button onPress={() => this.setModalVisible(false)} title="Done" />
                   </View>
                 </View>
               </View>
--- a/NativeApp/src/components/Common/CustomDatePickerIOS/index.js
+++ b/NativeApp/src/components/Common/CustomDatePickerIOS/index.js
@@ -1,29 +1,124 @@
-import React from 'react';
-import { View, DatePickerIOS } from 'react-native';
+import React, { Component, Fragment } from 'react';
+import {
+  View,
+  Text,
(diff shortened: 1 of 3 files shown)


NativeApp/src/components/Common/CustomDatePickerIOS/index.js
```diff
@@ -1,29 +1,122 @@
...
-};
+class CustomDatePickerIOS extends Component {
+  static propTypes = {
+    chosenDate: PT.string.isRequired,
+    setDate: PT.func.isRequired,
+    minimumDate: PT.string,
+  };
+
+  static defaultProps = {
+    minimumDate: '',
```
d1wilko: can we hardcode an actual minimum date here?

d1wilko (changes requested):
Just one question in the review - also do we need an android date picker?

NativeApp/src/components/Common/CustomDatePickerIOS/index.js
kirsty-unosquare: We can, was previously set as `new Date()`. But the minimumDate prop was changed to receive a string instead. Changed it to be an empty string to the stop the prop type errors. Just forgot to add a value to it.

Was thinking of setting the default using `moment().format('YYYY-MM-DD')`. Therefore the current date is set as the default and the end picker can never be set to the day before. Unless the start date is changed.

kirsty-unosquare:
@d1wilko Android picker is being made on another branch. Requires slightly different set up. So thought it was best to separate it from the iOS version.

NativeApp/src/components/Common/CustomDatePickerIOS/index.js
kirsty-unosquare: Same idea with the start date picker.

NativeApp/src/components/Common/CustomDatePickerIOS/index.js
```diff
@@ -19,7 +19,7 @@ class CustomDatePickerIOS extends Component {
   };
 
   static defaultProps = {
-    minimumDate: '',
+    minimumDate: moment().format('YYYY-MM-DD'),
```
d1wilko: am I right in saying that moment will return the current date here? - if so we will not be able to retroactively book days off, is this expected behaviour?

d1wilko (approved):
one comment - but approving

NativeApp/src/components/Common/CustomDatePickerIOS/index.js
kirsty-unosquare: Yep, it will set it to the current date. Technically you shouldn't be able to book a holiday that's already in the past. Couldn't think of any use case for needing to access a previous date when booking holidays.

d1wilko: in my previous experience with Liberty's system you could book days retroactively, e.g. if you last minute booked off a day and needed to book it into the system after you come back - but that's a decision for the BA i reckon

kirsty-unosquare: Didn't think of that. I was just basing it on having to book holidays in advance. It's an easy change if it needs updated, so gonna merge this. :)

PR #76, opened 2018-07-10
Task/s1.5.9 events and event type

peter-drysdale:
Updated schema to generate event table instead of holiday table, event type table, changed entity and dto to have Event/EventDto name and added EventType and EventTypeId respectively. Updated repository, services and controllers to use "/events/" url and tested saving, retrieving and updating. All API calls working normally, mapping configs updated to reflect new changes, unit tests renamed.


BrownMark:
So in don't think the consumers should be aware of the underlying schema. Can we please keep the holiday controller in place and keep the consumers abstracted?

BrownMark (changes requested):
Let's keep the holiday controller in place

BrownMark (approved):

PR #78, opened 2018-07-11
Feat/date picker android

kirsty-unosquare:
Converted the custom date picker from just iOS to being compatible with both android and iOS. Also removing the .env file.


d1wilko (approved):
no issues for me

William-Unosquare (approved):
Looks good

PR #82, opened 2018-07-12
extracted login form from login page

paulmatchett:


ghost (approved):
Nice, looks a lot cleaner to me. I'd wait for feedback from the others as well.

ghost (commented):
Reverting my approval, until I see the finished product

ghost (changes requested):
changed

web-app/src/components/LoginForm/index.jsx
```diff
@@ -0,0 +1,62 @@
...
+          required: true,
+          isEmail: true,
+        }}
+      />
+      <Input
+        type="input"
+        htmlAttrs={{
+          type: 'password',
+          name: 'password',
+          placeholder: 'Enter an password',
```
d1wilko: I think this should be 'Enter **a** password'

web-app/src/components/LoginForm/container.js
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import Swal from 'sweetalert2';
+import { PropTypes as PT } from 'prop-types';
+import { userLogin } from '../../services/userService';
+
+export default Wrapped =>
```
d1wilko, line 6: Can we name this something that better explains what it is? `Wrapped` is a bit abstract

d1wilko (changes requested):
Some real good stuff, 2 suggestions

web-app/src/components/LoginForm/container.js
paulmatchett: Wrapped is what has been used everywhere in the project that has this pattern. I was just following the convention.

d1wilko: fair enough, not a big fan of it but it's cool

web-app/src/components/LoginForm/index.jsx
paulmatchett: typo corrected

web-app/src/components/LoginForm/container.js
paulmatchett: I agree. I'll put it to the group and see if we can make it a fix throughout the project on another pull request.

web-app/src/components/LoginForm/index.jsx
d1wilko: can you push the change?

d1wilko (approved):
looks good to me

BrownMark (approved):

PR #83, opened 2018-07-12
Feat/s1.6.4 edit leave request

kirsty-unosquare:
App now allows you to update booked holidays.
 - new PUT request for updating holidays
 - booking submit button displays either request or update
 - formatting taken holidays now includes the holiday id

### DIFF
--- a/NativeApp/src/components/Home/container.js
+++ b/NativeApp/src/components/Home/container.js
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import { PropTypes as PT } from 'prop-types';
 import moment from 'moment';
-import _ from 'lodash';
+import { has, get } from 'lodash';
 import { userLogout } from '../../utilities/currentUser';
 import { getTakenHolidays } from '../../utilities/holidays';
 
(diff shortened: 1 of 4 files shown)


NativeApp/src/components/Home/container.js
```diff
@@ -43,14 +39,12 @@ export default Container => class extends Component {
...
       if (day) {
-        this.setState({
-          booking: {
-            startDate: day.dateString,
-            endDate: day.dateString,
-          },
-        });
-        navigation.push('Booking', { date: day.dateString });
+        const booked = has(takenHolidays, day.dateString);
+        const holiday = get(takenHolidays, day.dateString, 0);
```
William-Unosquare, line 46: Is there any way you can use the value returned from `const holiday = get(takenHolidays, day.dateString, 0);` to find out `booked`, rather than making two searches through the takenHolidays object?

William-Unosquare (approved):
Looks good, one wee question but I'm still gonna approve.

NativeApp/src/components/Home/container.js
kirsty-unosquare: Not 100% sure, for the dates are set as keys within the `takenHoliday` object. The holiday returns an object containing details from the matching selected date, so for example `{textColor: "white", color: "#ff9b34", holId: 1}`.

Booked looks to see if the `takenHolidays` object contains the selected date. It returns a boolean.

I need both values passed down, this was the best way I could think of doing it.

kirsty-unosquare: Example of `takenHolidays` object.
`2018-07-01: {textColor: "white", color: "#ff9b34", holId: 1}`

William-Unosquare: I see. I thought something like this might be the case which is why I still approved 👍

PR #87, opened 2018-07-17
Beginnings of ability to add a contract

William-Unosquare:
This feature is only partially complete, but I'm now blocked from continuing this because I need the backend to be developed more in this area.

Despite being partially complete it's already a big PR, so here's an overview:
- Adds a `Steps` component that renders different content based on the value of a step prop. Useful for splitting up UI flows. I want to extend this later to have breadcrumbs.
- Creating a contract will be split up into 3 steps:
  - ✅ Search and select the user the contract concerns, which is done in this PR.
  - 🔴 Search for a client, and select a team attached to the client that the contract concerns.
  - ✅ Enter the start and end dates of the contract, which is mostly done in this PR.


paulmatchett (approved):
Looks good to me

PR #88, opened 2018-07-17
Update/client section

paulmatchett:


web-app/src/components/CreateClientForm/index.jsx
```diff
@@ -0,0 +1,154 @@
...
+            placeholder: 'Enter an Contact Name',
+          }}
+          value={formData.contactName}
+          label="Contact Name:"
+          rules={{
+            required: true,
+          }}
+        />
+        <Input
+          type="select"
```
William-Unosquare: Not sure a select box is the most optimal thing in terms of UX for this type of input

William-Unosquare (changes requested):
Looks good. As we talked about we've been using `react-table` for the tables, generally just for consistency across the board, which makes your table component a bit outdated before it's even merged in.  But I think if this works for now then keep it as is until we get other features done. There was just one other thing I spotted.

web-app/src/components/CreateClientForm/index.jsx
paulmatchett: Updated this to a input field with a type of number

William-Unosquare (approved):

PR #89, opened 2018-07-17
Display user details

kirsty-unosquare:
Now able to see user details under the user tab. No styling just logic being passed through so Lee can get started on the styles while I'm off.


d1wilko (approved):
no concerns here

PR #93, opened 2018-07-20
Create Contract

William-Unosquare:
Completes the 3 step process of creating a contract. The process is as follows:
- Search and select a user
- Pick a client's team
- Pick the start and end date

### DIFF
--- a/web-app/src/services/clientService.js
+++ b/web-app/src/services/clientService.js
@@ -7,3 +7,7 @@ export const createClient = client => {
 export const getAllClients = () => {
   return axios.get('/clients/');
 };
+
+export const getClientById = clientId => {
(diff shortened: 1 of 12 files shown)


web-app/src/services/clientService.js
```diff
@@ -7,27 +7,3 @@ export const createClient = client => {
 export const getAllClients = () => {
   return axios.get('/clients/');
 };
-
-export const getClientById = clientId => {
-  return axios.get(`/clients/${clientId}`);
```
paulmatchett: getClientById is being used for updating a client?

William-Unosquare: Well spotted, sorry!

paulmatchett (approved):

PR #94, opened 2018-07-23
View Contracts

William-Unosquare:
You can now view contracts by searching for an employee's full name. Right now it only shows start and end dates as we get no info regarding team names / client names from the back-end.


paulmatchett (approved):
That all looks nice a tidy to me

PR #96, opened 2018-07-23
Fix updating bug on user info.

kirsty-unosquare:
Just a teeny weeny fix for user tab not updating. I did noticed tabs were re rendering in the background on tab change. I think this is due to the navigation state changing therefore re rendering all of the tabs.

I've stopped running any calls in the background by adding an event listener `didfocus` in the mean time.

### DIFF
--- a/NativeApp/src/components/User/container.js
+++ b/NativeApp/src/components/User/container.js
@@ -1,8 +1,19 @@
 import React, { Component } from 'react';
+import { PropTypes as PT } from 'prop-types';
 import { getTakenHolidays, getRemainingHolidays } from '../../utilities/holidays';
 import { userProfile } from '../../utilities/currentUser';
 
 export default Container => class extends Component {
+  static propTypes = {
+    navigation: PT.shape({
+      navigate: PT.func,
+    }),
+  }
+
+  static defaultProps = {
+    navigation: {},
+  }
+
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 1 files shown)


NativeApp/src/components/User/container.js
```diff
@@ -17,21 +28,29 @@ export default Container => class extends Component {
...
+      getTakenHolidays()
+        .then(takenHolidays => this.setState({ takenHolidays }));
+
+      getRemainingHolidays()
+        .then(remainingHolidays => this.setState({ remainingHolidays }));
+
+      userProfile()
+        .then(employee => this.setState(
+          {
+            employee: {
```
d1wilko: would this not work as `{ employee }`?

d1wilko (changes requested):
looks good, one question

NativeApp/src/components/User/container.js
kirsty-unosquare: You are correct. Good spot 👍 .

NativeApp/src/components/User/container.js
```diff
@@ -17,21 +28,22 @@ export default Container => class extends Component {
...
+
+    this.sub = navigation.addListener('didFocus', () => {
+      getTakenHolidays()
+        .then(takenHolidays => this.setState({ takenHolidays }));
+
+      getRemainingHolidays()
+        .then(remainingHolidays => this.setState({ remainingHolidays }));
+
+      userProfile()
+        .then(employee => this.setState({ employee }));
```
William-Unosquare, line 41: Still gonna approve, but it may be good to get some error handling on these calls

William-Unosquare (approved):

d1wilko (approved):
looks good - I agree with Will I Am but going to approve

NativeApp/src/components/User/container.js
kirsty-unosquare: I do have some error handling set within the service calls ( displays an alert ). @William-Unosquare Do think it would be better to set them within the container files or keep them within services or utilities folder?

William-Unosquare: @kirsty-unosquare That's a cool idea. I think that depends on if your setup makes it easy or not to add additional logic in specific cases. (For example, navigate away from a screen on an error). Sounds absolutely fine to me for right now though, so again, approving 👍

kirsty-unosquare: Sweet gonna merge this branch.

PR #101, opened 2018-07-25
Refinements

William-Unosquare:
This is a hefty PR in terms of files changed, but most changes are small improvements that should help improve the experience or improve the code. I've kept away from large logic changes.

Here's a list of changes:
- When creating something such as a user, team, or contract as an admin, when the creation was a success a toast message will appear and reset the form.
  - A Toast function can be found in `utilities/Notifications.js` with some default settings
- Forms are moved into the directory where their parent components are, to clean up the file structure.
- Admin routes are now much cleaner and follow the following convention: 
  - View all of something: `/admin/something`
  - Create something: `/admin/something/new`
  - Update something: `/admin/something/1`
- A welcome toast will appear when you visit the site / log in.
- Log out modal is fixed now that it uses the `SweetAlert2` API as it's options instead of `SweetAlert`'s API.
- We're now using [Rodal](https://github.com/chenjiahan/rodal) for our modals. This is just so we have a solid modal component being used and it brings extras such as clicking the background to dismiss, and animations. Our modal component has been updated to use this.


paulmatchett (approved):
nice idea to implement toast. Just ties up all the success event with a similar feel

PR #109, opened 2018-07-26
Simplified create client form to match api

William-Unosquare:
- Updates create client page to match the API (only need to supply a name now)
- Cleaned the code a bit. Moved service calls into the page's container, out of the form's container. The way I see it is the form container should just manage the form itself.
- Toast notification for success, alert notification for errors.


William-Unosquare:
Don't merge this until #110 is merged into this branch.

BrownMark (approved):

PR #110, opened 2018-07-26
[Merge before #109] Updates View All Clients screen

William-Unosquare:
Merge this before #109 is merged.

All clients screen now uses react table. You can select a row and a modal will appear that we can fill in with information later, but right now links to the ability to update a client.


BrownMark (approved):
Looks good man

PR #112, opened 2018-07-27
Fix Pending Holidays

William-Unosquare:
- Fixes issue where pending holidays weren't being rendered in the table.
- Adds new getHolidayByStatus call, and the pending holiday's page now uses this instead of getAllHolidays.
- Removes non-existant endpoint in holiday service.


paulmatchett (approved):
looks good to me

BrownMark (approved):

PR #113, opened 2018-07-27
Request half day.

kirsty-unosquare:
Requesting half day now available within the booking screen. Style updates.

new object to retrieve holiday status colour.


NativeApp/src/components/Booking/View/index.js
```diff
@@ -17,37 +19,93 @@ const BookingView = (props) => {
...
+    status,
   } = props;
 
   const renderButton = booked
     ? (
-      <Fragment>
-        <Button onPress={() => updateHoliday(false)} title="Update Holiday" />
-        <Button onPress={() => updateHoliday(true)} title="Cancel Holiday" />
-      </Fragment>
+      <View style={styles.buttonContainer}>
```
bgribz: Just a thought, but would this be cleaner kept as its own component? Just seems to be quite a chunky ternary.

NativeApp/src/components/Booking/View/index.js
```diff
@@ -17,37 +19,93 @@ const BookingView = (props) => {
...
+    : (
+      <Button
+        onPress={submitRequest}
+        title="Request"
+        backgroundColor="#00DCFA"
+        borderRadius={5}
+      />
+    );
 
+  const renderStatusBar = booked
```
bgribz: Again, having this as its own component would allow for `booked && <StatusBar />` in the render rather than declaring this const.

NativeApp/src/components/Booking/container.js
```diff
@@ -126,14 +132,31 @@ export default Container => class extends Component {
...
 
     return (
       <Container
         startDate={booking.startDate}
         endDate={booking.endDate}
+        halfDay={booking.halfDay}
+        updateHalfDay={this.updateHalfDay}
         booked={booked}
+        statusId={booking.statusId}
+        status={booking.status}
```
William-Unosquare: Would be cleaner to just pass through a `booking` prop as an object here, instead of splitting it all up like this.

NativeApp/src/components/Booking/View/index.js
```diff
@@ -59,16 +117,15 @@ BookingView.propTypes = {
...
-    flex: 1,
-    justifyContent: 'flex-start',
-    alignItems: 'center',
-    paddingTop: 50,
-    backgroundColor: '#fff',
-  },
-});
+BookingView.defaultProps = {
+  status: '',
+  statusId: 0,
```
William-Unosquare: Not sure there's any need for these defaultProps if this info is coming from the default state in the booking container, because the default state is set up the same way. But see my other comment on modifying all these props.

bgribz (changes requested):

William-Unosquare (changes requested):
Just one bit of cleanup I think you should do. If you disagree then lemme know and I'll approve.

NativeApp/src/components/Booking/container.js
kirsty-unosquare: Totally agree, did the same for `holiday` within the home container. Forgot about this one.

kirsty-unosquare:
@William-Unosquare @bgribz Should be all good now...hopefully.

NativeApp/src/components/Booking/StatusBar/index.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { View, Text, StyleSheet } from 'react-native';
+import { PropTypes as PT } from 'prop-types';
+import holidayStatusColor from '../../../utilities/holidayStatus';
+
+const StatusBar = (props) => {
```
William-Unosquare:
```
const StatusBar = ({booking}) => {
const {destructure, booking, props} = booking;
```

Would clean this file up a good bit

NativeApp/src/components/Booking/container.js
```diff
@@ -126,14 +132,27 @@ export default Container => class extends Component {
...
+      booking: {
+        ...booking,
+        endDate: startDate,
+        halfDay: !halfDay,
+      },
+    });
+  }
+
   render() {
     const { booking, booked } = this.state;
```
William-Unosquare, line 149: Destructure booking here too. Normally I don't mind this but you've done it in `updateHalfDay` just above.

William-Unosquare (approved):
Approving but there are a couple more bits you could clean up 😄

NativeApp/src/components/Booking/container.js
kirsty-unosquare: Don't think there would be much point destructuring booking. It's being passed to the View, instead of passing the values individually.

bgribz (approved):

PR #114, opened 2018-07-27
Holiday list fix

William-Unosquare:
- Holiday list status now displays, and changed styling to be less harsh.
- The requested column now actually shows the correct info.
  - As a bonus feature, if the holiday was requested 5 or more days ago, and it's still pending, the text is highlighted in bold red to attract attention.

There are some features removed here, but this is because this will be moving in the direction of a modal appearing on a row click instead of action buttons being inline.


web-app/src/components/HolidayList/index.jsx
```diff
@@ -1,86 +1,54 @@
...
+  const { holidays, columns, onRowClick } = props;
+  return !holidays || holidays.length === 0 ? (
+    <p>There are no holidays to show</p>
+  ) : (
+    renderTable(holidays, columns, onRowClick)
   );
 };
 
 HolidayList.propTypes = {
   holidays: PT.array,
```
paulmatchett: Should there be a default value for holidays if it is no required?

paulmatchett (approved):
nice small PR man. Looks good to me

web-app/src/components/HolidayList/index.jsx
William-Unosquare: Good spot - the holidays should be required.

PR #116, opened 2018-07-30
Small fix for loading user info.

kirsty-unosquare:
Super tiny fix to update holiday status type to match new schema.


William-Unosquare (approved):
👍

PR #119, opened 2018-07-30
Feat/holiday modal

William-Unosquare:
Adds a holiday modal to the pending holidays page, but can be applied to anywhere else in the app that displays a holiday table.

It allows the admin to approve or reject holidays but needs to be updated to implement the actual calls to the backend once they're merged.

Updated button components to accept an array of labels. If the button is clicked on when multiple labels are specified it will display each label, then call the action when all labels have been shown. A useful case for this is:

`['Reject Holiday', 'Are you sure?']` for a confirmation button that requires 2 clicks


web-app/src/components/HolidayModal/container.js
```diff
@@ -0,0 +1,36 @@
...
+    rejectHoliday = holidayId => {
+      console.log('rejected');
+    };
+
+    render() {
+      const { holiday } = this.props;
+      if (isEmpty(holiday)) return null;
+      return (
+        <Wrapped
+          {...this.props}
```
paulmatchett: I'm not a fan of using the spread operator for passing this.props and this.state down as props. Easier to read if explicit?

web-app/src/components/PendingHolidays/container.js
```diff
@@ -19,9 +19,15 @@ export default Wrapped =>
...
 
+    selectHoliday = holiday => this.setState({ selectedHoliday: holiday });
+
     render() {
       return (
-        <Wrapped {...this.props} pendingHolidays={this.state.pendingHolidays} />
+        <Wrapped
+          {...this.props}
+          {...this.state}
+          selectHoliday={this.selectHoliday}
```
paulmatchett, line 29: same as before. A lot easier being to read/understand flow when being explicit with what is being passed in?

bgribz:
I'm not dying about the multi label feature for buttons. 

I'm not sure there's enough scenarios to warrant adding this complexity to something that should be kept quite simple, and even in the scenarios where it could be used, it doesn't do a very good job - e.g. changing the label to "Are you sure?" doesn't actually present the user with any new options and is probably bad UX. 

In my opinion we should have an additional modal for action confirmation, or use something like SweetAlert (if that's still in the app).

Thoughts?

William-Unosquare:
@bgribz I'd normally agree. Normally models are better for this job. But we're now using a lot of modals in the app and I'd rather not have modals on top of modals. 

I see your concerns that it's turning a commonly used component into something more complex, but the component's flow is still kept just as simple as before if you're not specifying an array.

Some more opinions would be good but let's not get held up on this with the presentation being tomorrow.

**Edit:**
After thinking about this a bit more I'm gonna remove this in favour of adding an undo feature.

web-app/src/components/HolidayModal/container.js
William-Unosquare: Good spot. Agreed.

paulmatchett (approved):

PR #120, opened 2018-07-30
Improve logging out feature

kirsty-unosquare:
I've moved the log out button from the home screen into it's own tab. On tab click an alert appears declaring if the user wants to log out.

Also commented out the plus icon on home screen for the time being.

### DIFF
--- a/NativeApp/src/screens/Logout.js
+++ b/NativeApp/src/screens/Logout.js
@@ -40,7 +40,7 @@ export default class LogoutScreen extends Component {
   handleLogout = () => {
     const { navigation } = this.props;
     userLogout()
-      .then(navigation.navigate('Auth'));
+      .then(() => navigation.navigate('Auth'));
   }
 
   render() {
--- a/NativeApp/src/screens/Logout.js
+++ b/NativeApp/src/screens/Logout.js
@@ -1,0 +1,49 @@
+import React, { Component } from 'react';
+import { View, Alert } from 'react-native';
+import { PropTypes as PT } from 'prop-types';
+import Icon from 'react-native-vector-icons/FontAwesome';
+import { userLogout } from '../utilities/currentUser';
+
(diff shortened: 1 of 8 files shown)


William-Unosquare (approved):
The code looks clean and good to me.

NativeApp/src/screens/Logout.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  componentWillUnmount() {
+    this.sub.remove();
+  }
+
+  handleLogout = () => {
+    const { navigation } = this.props;
+    userLogout()
+      .then(navigation.navigate('Auth'));
```
bgribz: I'm not 100% on this, but I think this might be firing immediately rather than after a successful `userLogout`. I would have thought it needed to be `.then(() => navigation.navigate('Auth'))`

bgribz (changes requested):
Looks good, just a quick query.

NativeApp/src/screens/Logout.js
kirsty-unosquare: Haven't came across any issues or errors that would indicate it firing before login. Always have been able to successfully input login details before `navigate` is called. Think it would be safer including `.then(() => navigation.navigate('Auth'))` @bgribz ?

bgribz: In my head, as soon as `handleLogout` is called, it's evaluating the `navigation.navigate('Auth')` call immediately and passing that to `then`, whereas we want to pass a function that performs the navigation.

It'd be interesting to stick some logging in and see the order in which things are firing, but if you don't have time I say include `.then(() => navigation.navigate('Auth'))` just to be safe 👍

kirsty-unosquare: Actually @bgribz your right. The navigation was being called first. Good spot! 👍

bgribz (approved):

PR #123, opened 2018-07-30
Add holiday modal to all other pages with a holiday list

William-Unosquare:
Adds the holiday modal to all other pages, even ones outside of the admin area. I'm going to add conditional checks to remove the approve/reject buttons if they're not an admin or team-leader.


paulmatchett (approved):
looks good. Would tidy the ...the,state and ...this.props going into some wrapped components, but other than that.

PR #125, opened 2018-07-30
Fix/requested dates

William-Unosquare:
- Fixes issue where the schema update broke the "days until" logic.


BrownMark (approved):
Looks good, using the daterCreated value correctly.

PR #134, opened 2018-07-31
Fix/holiday update request - mobile

kirsty-unosquare:
Fixes update holidays and adds new cancel request for mobile.

### DIFF
--- a/NativeApp/src/components/Booking/RequestButton/index.js
+++ b/NativeApp/src/components/Booking/RequestButton/index.js
@@ -4,19 +4,19 @@ import { Button } from 'react-native-elements';
 import { PropTypes as PT } from 'prop-types';
 
 const RequestButton = (props) => {
-  const { updateHoliday, booked, submitRequest } = props;
+  const { updateHoliday, cancelHoliday, booked, submitRequest } = props;
 
   return (
     booked ? (
       <View>
         <Button
-          onPress={() => updateHoliday(false)}
+          onPress={() => updateHoliday()}
           title="Update Holiday"
           backgroundColor="#00DCFA"
           borderRadius={5}
         />
         <Button
-          onPress={() => updateHoliday(true)}
+          onPress={() => cancelHoliday()}
           title="Cancel Holiday"
           containerViewStyle={{ marginTop: 20 }}
           borderRadius={5}
@@ -37,6 +37,7 @@ export default RequestButton;
 
 RequestButton.propTypes = {
   updateHoliday: PT.func.isRequired,
+  cancelHoliday: PT.func.isRequired,
   booked: PT.bool.isRequired,
   submitRequest: PT.func.isRequired,
 };
--- a/NativeApp/src/components/Booking/RequestButton/index.js
+++ b/NativeApp/src/components/Booking/RequestButton/index.js
@@ -4,7 +4,7 @@ import { Button } from 'react-native-elements';
 import { PropTypes as PT } from 'prop-types';
 
 const RequestButton = (props) => {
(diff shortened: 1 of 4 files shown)


NativeApp/src/components/Booking/RequestButton/index.js
```diff
@@ -10,15 +10,15 @@ const RequestButton = (props) => {
     booked ? (
       <View>
         <Button
-          onPress={() => updateHoliday(false)}
+          onPress={() => updateHoliday()}
```
bgribz: This can just be `onPress={updateHoliday}`

NativeApp/src/components/Booking/RequestButton/index.js
```diff
@@ -10,15 +10,15 @@ const RequestButton = (props) => {
...
           title="Update Holiday"
           backgroundColor="#00DCFA"
           borderRadius={5}
           loading={loading}
           loadingRight
         />
         <Button
-          onPress={() => updateHoliday(true)}
+          onPress={() => cancelHoliday()}
```
bgribz: onPress={cancelHoliday}

NativeApp/src/components/Booking/container.js
```diff
@@ -115,17 +115,16 @@ export default Container => class extends Component {
       });
   }
 
-  updateHoliday = (cancel) => {
-    const { booking, user } = this.state;
+  updateHoliday = () => {
+    const { booking } = this.state;
+    const { endDate, halfDay, startDate, holId } = booking;
```
bgribz: You can destructure this on one line

`const { booking: { endDate, halfDay, startDate, holId } } = this.state;`

NativeApp/src/components/Booking/container.js
```diff
@@ -138,6 +137,25 @@ export default Container => class extends Component {
...
+    const { booking } = this.state;
+    const { holId } = booking;
+    const { navigation } = this.props;
+
+    const request = {
+      holidayId: holId,
+    };
+
+    cancelHolidayRequest(request)
+      .then(() => {
```
bgribz: I think (if you wanted) you could have `.then(navigation.pop)` here

bgribz (changes requested):
All looks swell, just a few tidy ups but it's mainly down to personal preference!

NativeApp/src/components/Booking/container.js
kirsty-unosquare: Would it not get called before? Like it was on login. @bgribz

bgribz: Nope, so

`.then(navigation.pop)` is passing along a reference to the pop function within the navigation object 

`.then(() => navigation.pop())` is passing along an anonymous function which will call the pop function within the navigation object. This is more or less the same as above (I think), only slightly less cleaner.

`.then(navigation.pop())` would immediately call the pop function within the navigation object, and pass whatever it gets back to `then`, which isn't what we want here.

Hope that clears it up a bit!

kirsty-unosquare: oooo got it now. I didn't see the `()` that was missing from what you suggested.

kirsty-unosquare: Actually just tried it. Doesn't fire the `pop` action.

kirsty-unosquare: @bgribz Had to revert it back. Seems `() => navigation.pop()` is the only way it will work.

kirsty-unosquare:
@bgribz Have those changes up.

bgribz (approved):

PR #139, opened 2018-08-02
Fix table sorting and filtering

William-Unosquare:
Finally got my head around how `react-table` works, so a few improvements:

**Holiday Tables**
- Can now be sorted and filtered by status.
- Dates can now be sorted properly.
- New filtering logic for dates, as the sorting broke the default for them.
- Sorting on the tables are now sorted by the earliest start date. *This means a start date is always required*

**All Tables**
- All tables now take a `pageSize` prop, which changes how many rows are shown.
- All admin pages now have a default of 20 rows.

### DIFF
--- a/web-app/src/components/AllClients/index.jsx
+++ b/web-app/src/components/AllClients/index.jsx
@@ -30,6 +30,7 @@ export const AllClients = ({
         clients={clients}
         columns={['clientName']}
         onRowClick={data => selectClient(data)}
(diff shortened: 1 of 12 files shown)


kirsty-unosquare:
Had a look through and looks good to me. But I think you should get a second pair of eyes on it. Not sure how `react-table` works.

BrownMark (approved):

PR #141, opened 2018-08-02
Add additional navigation to admin employee pages

William-Unosquare:
- Link to All Employees from Create Employee.
- Link to Create Employee from All Employees.


BrownMark (approved):
Looks good Will.

PR #143, opened 2018-08-03
Update/dashboard redux

paulmatchett:
redux store implemented into dashboard


William-Unosquare:
I'm going to work on this a bit on Monday.

We shouldn't merge in two different styles of reducer so I'm going to update it to the style already used. I'll also add some selectors in.

William-Unosquare:
I looked at this more closely this morning. There's a lot that needs to be changed. I'm going to say we don't merge this in right now and Kirsty had the same opinion. 

I was hoping I'd get it fixed up this morning but the reason I'm not going to do that is that I'm not sure if there are reasons behind some of the stuff @paul-matchett has done here. And there really is a lot to change. I'd much rather pair up with Paul and go through it with him.

I think the main area that needs to be worked on here is the actions & action creators. There seems to be an extra layer added into all this logic that the action creators should be handling. For example you have all this in the actions file: 

``` javascript
export const showBookingModal = () => {
  return {
    type: actionTypes.SHOW_BOOKING_MODAL,
    bookingModalOpen: true,
  };
};

 export const hideBookingModal = () => {
  return {
    type: actionTypes.HIDE_BOOKING_MODAL,
    bookingModalOpen: false,
  };
};
```

``` javascript
export const toggleBookingModal = bookingModalOpen => {
  return dispatch => {
    if (bookingModalOpen) {
      dispatch(showBookingModal());
    } else {
      dispatch(hideBookingModal());
    }
  };
};
```

When this would be enough:

``` javascript
 export const toggleBookingModal = shouldDisplay => {
  return {
    type: TOGGLE_BOOKING_MODAL,
    payload: shouldDisplay,
  };
};
```

You should then dispatch `toggleBookingModal` elsewhere in the code, where you'd be toggling the modal.

William-Unosquare (changes requested):
See above ^

William-Unosquare (approved):

William-Unosquare:
This is hopefully all good to merge now after refactoring, need someone else to review

web-app/src/actions/dashboard.js
```diff
@@ -0,0 +1,76 @@
...
+};
+
+export const updateEventDuration = event => {
+  return {
+    type: actionTypes.UPDATE_EVENT_DURATION,
+    payload: getEventDuration(event),
+  };
+};
+
+export const toggleBookingModal = open => {
```
kirsty-unosquare, line 26: Not keen on the payload name `open`. If it's toggle then it could also be closing the modal. Maybe name it `toggle` or something similar. :)

web-app/src/components/BookingModal/BookingModalForm/container.js
```diff
@@ -1,35 +1,58 @@
...
+          eventTypeId: 1,
+          isHalfday: false,
+          start: moment(),
         },
         formIsValid: true,
       };
     }
 
+    componentDidMount = () => {
+      this.setState({
```
kirsty-unosquare: Just my own thought. But seeing a lot of `this.props...`. I think it would be neater to do ` const { booking: { halfDay, start, eventType }} = this.props;`

web-app/src/components/BookingModal/container.js
```diff
@@ -0,0 +1,63 @@
...
+  return {
+    booking: getBooking(state),
+    bookingModalOpen: bookingModalOpen(state),
+    bookingDuration: getBookingDuration(state),
+  };
+};
+
+const mapDispatchToProps = dispatch => {
+  return {
+    toggleBookingModal: open => dispatch(toggleBookingModal(open)),
```
kirsty-unosquare, line 59: Same here. `open` doesn't make sense when your toggling a modal. In this case your closing it.

web-app/src/components/Legend/container.js
```diff
@@ -74,8 +74,15 @@ const LengendContainer = Wrapped =>
       };
     }
 
-    componentWillMount = () => {
-      this.createEmployeesList();
+    componentWillUpdate = (nextProps, nextState) => {
```
kirsty-unosquare: Just a heads up react now declare `componentWillUpdate` as unsafe. From the top of my head I think they prefer to use `componentDidUpdate`.
https://reactjs.org/docs/react-component.html#unsafe_componentwillupdate

web-app/src/pages/Dashboard/container.js
```diff
@@ -1,197 +1,98 @@
...
-          this.setTakenHolidayState(response.data);
-        })
-        .catch(error => {
-          Swal({
-            text: error.message,
-            title: 'Could not get taken holidays',
-            type: 'error',
-          });
-        });
+    componentWillUpdate = (nextProps, nextState) => {
```
kirsty-unosquare: Same here

kirsty-unosquare (commented):
Had a look through, just a couple of tidy ups. A lot to look over so wouldn't mind someone else reviewing it.

web-app/src/components/BookingModal/container.js
William-Unosquare: To me, it makes more sense when considered as a bool. `open: true`, `open: false`.  

`toggle: true`, `toggle: false` sounds less descriptive.  Lemme know what you think.

PR #146, opened 2018-08-06
Error handling & Loading handling when retrieving user profile

William-Unosquare:
- Errors are handled correctly for retrieving the user's profile. Previously a modal would appear and contents of pages would still render, risking crashes. Now the page render is blocked and an error is shown on the page itself.
- A [loader ](http://tobiasahlin.com/spinkit/)is now displayed while the user's profile is being retrieved.


kirsty-unosquare (approved):
Looks good. :)

PR #147, opened 2018-08-07
Add new filter to other list components

William-Unosquare:
Adds the same filter that the holiday list uses into the rest of the lists (users, contracts, teams, clients).

This is a pretty small PR that looks bigger because I had to change stateless components into ones with the state.


kirsty-unosquare (approved):
Looks good.

web-app/src/components/ClientList/index.jsx
```diff
@@ -1,56 +1,81 @@
...
+  render() {
+    const { clients, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    return !clients || clients.length === 0 ? (
```
d1wilko: the logic seem fine, but a ternary operator over multiple lines? 🤢

use an if else instead

web-app/src/components/ContractList/index.jsx
```diff
@@ -1,55 +1,81 @@
...
+  render() {
+    const { contracts, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    return !contracts || contracts.length === 0 ? (
```
d1wilko: again ternary 🤢

web-app/src/components/ContractList/index.jsx
```diff
@@ -1,55 +1,81 @@
...
-  columns: PT.array.isRequired,
-  onRowClick: PT.func,
-  pageSize: PT.number,
-};
-
-ContractList.defaultProps = {
-  onRowClick: () => {},
-  pageSize: 10,
-};
+class ContractList extends Component {
```
d1wilko, line 7: the vast majority of this code seems to have been copied and pasted from clientList... is there no way we can centralise this code?

web-app/src/components/TeamList/index.jsx
```diff
@@ -1,55 +1,81 @@
...
-  columns: PT.array.isRequired,
-  onRowClick: PT.func,
-  pageSize: PT.number,
-};
-
-TeamList.defaultProps = {
-  onRowClick: () => {},
-  pageSize: 10,
-};
+class TeamList extends Component {
```
d1wilko, line 7: now I'm more certain about not repeating the code

web-app/src/components/TeamList/index.jsx
```diff
@@ -1,55 +1,81 @@
...
+  render() {
+    const { teams, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    return !teams || teams.length === 0 ? (
```
d1wilko: 🤢

web-app/src/components/UserList/index.jsx
```diff
@@ -1,60 +1,81 @@
...
-  pageSize: PT.number,
-};
-
-UserList.defaultProps = {
-  onRowClick: () => {},
-  pageSize: 10,
-};
-
-export default container(UserList);
+class UserList extends Component {
```
d1wilko, line 7: yep, that does it

web-app/src/components/UserList/index.jsx
```diff
@@ -1,60 +1,81 @@
...
+  render() {
+    const { users, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    return !users || users.length === 0 ? (
```
d1wilko: 💀

d1wilko (changes requested):
Few changes could/should be made here

web-app/src/components/UserList/index.jsx
William-Unosquare: Yeh, I know this code is repeated a good bit. It used to make sense to have them all separate but they've been refactored over time (and even in this PR) to make them more flexible. Believe me, I noticed when I was changing these that refactoring these into a single more flexible component may make sense to do.

I didn't want to make it the focus of this PR though because there are some considerations to take into place, like if in the future it will make sense to have these separate again. If this gets merged I'll work today on converting all this into a single component, just to keep these PR's clean.

web-app/src/components/UserList/index.jsx
William-Unosquare: Will fix 💯

web-app/src/components/ClientList/index.jsx
```diff
@@ -1,56 +1,83 @@
...
+  render() {
+    const { clients, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    if (clients.length > 0) {
```
d1wilko, line 66: based on your previous PR should this not be `if (clients && clients.length > 0)`?

I'm making the assumption that sometimes clients will be undefined from the previous code

web-app/src/components/ContractList/index.jsx
```diff
@@ -1,55 +1,83 @@
...
+  render() {
+    const { contracts, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    if (contracts.length > 0) {
```
d1wilko, line 66: same as above

web-app/src/components/TeamList/index.jsx
```diff
@@ -1,55 +1,83 @@
...
+  render() {
+    const { teams, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    if (teams.length > 0) {
```
d1wilko, line 66: same as above

web-app/src/components/UserList/index.jsx
```diff
@@ -1,60 +1,83 @@
...
+  render() {
+    const { users, columns, onRowClick, pageSize } = this.props;
+    const formattedColumns = this.buildColumns(columns);
+
+    const labels = formattedColumns.reduce((acc, column) => {
+      acc.push(column.Header);
+      return acc;
+    }, []);
+
+    if (users.length > 0) {
```
d1wilko, line 66: same as above

d1wilko (changes requested):
one potential issue

web-app/src/components/ClientList/index.jsx
William-Unosquare: Yea, I don't know know why both were being checked previously. The data should never be undefined, which is why I've changed it here.

d1wilko (approved):

PR #154, opened 2018-08-10
All list components refactored into one component

William-Unosquare:
Previously when we were making a table we would use one of these:
`HolidayList`, `ContractList`, `TeamList`, `UserList`, `ClientList` and so on. They were split up because of how they were previously built required this.

However, over some PRs they've been refactored down to the point where they were really all the same component but with different cells imported. (As @d1wilko pointed out in #147)

This PR updates this so they all work off a central `DataTable` component, where you import the cell data in as a prop along with the data itself, and the columns you want.


paulmatchett (approved):

web-app/src/pages/Profile/index.jsx
```diff
@@ -40,11 +41,12 @@ const Profile = props => {
...
           </div>
-          <HolidayList
-            holidays={userHolidays}
+          <DataTable
+            data={userHolidays}
+            cells={HolidayCells}
             columns={['status', 'startDate', 'endDate', 'requestedDate']}
             onRowClick={holiday => selectHoliday(holiday)}
-          />
+          />>
```
kirsty-unosquare: You've a extra `>` here.

kirsty-unosquare (approved):
Looks good apart from the typo. :)

web-app/src/pages/Profile/index.jsx
William-Unosquare: Well spotted

PR #159, opened 2018-08-14
Event type added into booking.

kirsty-unosquare:
Event type selection has now been added into the booking form. But not functionally working due to no data being passed in from the backend ( correct me if I'm wrong ).

The majority of this PR is styling updates. @lee-cheung I updated some of the global font sizes. They were a tad big. Was taking up a lot of room.

simulator screen shot - iphone x - 2018-08-14 at 16 31 14
simulator screen shot - iphone x - 2018-08-14 at 16 32 06


William-Unosquare (approved):
Code looks good, as do the buttons! :)

PR #160, opened 2018-08-15
Webapp unit tests tidy

paulmatchett:
This is mostly a tidy up so that running npm test shows results without console errors.

Updates to testing styled components:
- Button Styled Component - disabled styles updated from colour to opacity
- ErrorBox Styled Component - Imported Styled Theme to recognise prop colours
- Modal Styled Component - removed as component no longer exists

Updates to testing Services:
- holidayService - removed 'updateHolidays returns 200 if successful' as no longer relavant

Component test suites that are now being skipped:
- BookingModal Component
- CreateUser Container 
- CreateUser Component
- Login Container
- Input Component
- Form Component

Other changes included:
- Updating test components requiring props being passed in as jest.mocks
- sweetalert2 mock being updated to return a mock function within an object
- All snapshots have been updated


web-app/__tests__/components/BookingModal.js
```diff
@@ -42,17 +45,20 @@ describe('BookingModal', () => {
...
+        changeEnd={() => {}}
+        changeHalfday={() => {}}
+        requestHoliday={() => {}}
+        cancelHoliday={() => {}}
+        updateHoliday={() => {}}
+        changeWFH={() => {}}
+      />,
+    );
 
     expect(mockOnClick.mock.calls.length).toEqual(0);
```
William-Unosquare: Not sure this is really necessary.

William-Unosquare: Actually I think this whole area could just use:
`expect(mockOnClick).toBeCalled();`

https://jestjs.io/docs/en/mock-functions#custom-matchers

William-Unosquare (approved):
Looks good. It will put the tests in a better starting point when we come back to them.

PR #161, opened 2018-08-15
fixed cursor jumping bug in inputs

paulmatchett:
Changed Input class from React.Component to React.PureComponent as it implements shouldComponentUpdate() which was what was required to monitor props.value and prevent the cursor from jumping on change event.


William-Unosquare (approved):
What a fix! Nice one man.

PR #162, opened 2018-08-16
Fix period marker issue.

kirsty-unosquare:
Fixes a bug with period marker. Needed an extra check to make sure dates between start and end date don't use the single event maker.

### DIFF
--- a/NativeApp/src/components/Home/CustomDay/index.js
+++ b/NativeApp/src/components/Home/CustomDay/index.js
@@ -5,7 +5,15 @@ import Day from 'react-native-calendars/src/calendar/day/custom';
 import holidayStatusColor from '../../../utilities/holidayStatus';
 
 const CustomDay = (props) => {
-  const { marking: { halfDay, statusId, startingDate, endingDate } } = props;
+  const { marking:
+    {
+      halfDay,
+      statusId,
+      startingDate,
+      endingDate,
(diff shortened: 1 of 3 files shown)


William-Unosquare (approved):
👍

NativeApp/src/components/Home/CustomDay/index.js
```diff
@@ -28,19 +36,29 @@ const CustomDay = (props) => {
     );
   };
 
+  const renderFiller = (date) => {
+    let filler;
+    if (!sameDate) {
+      if (date || (!startingDate && !endingDate)) {
```
d1wilko, line 42: is there a need for 2 separate if statements? - not a huge deal, just might be cleaner as one

d1wilko (approved):
just one question, but approving anyway

NativeApp/src/components/Home/CustomDay/index.js
kirsty-unosquare: It was really for my own benefit, easier to read while I was working on it. But can change to a one liner. :)

William-Unosquare: @kirsty-unosquare I think it's easier to read this way aswell.

lee-cheung (approved):

PR #164, opened 2018-08-16
Refactor Booking Modal & Ability to create work from home events

William-Unosquare:
- Refactored `BookingModal` and `BookingModalForm`.
  - The main difference is all requests are now moved into the `BookingModal` container, for two reasons:
     1. Such important calls felt very hidden in the file structure.
     2. Cancel holiday call has nothing to do with the form, so it made sense to have all these requests together in the `BookingModal` container
  - General code clean up, removing duplicated code that wasn't required.
- You can now create a work from home event. Note it doesn't display on the calendar yet.
- Moved and restyled cancel event button so it's not mistaken as a close modal button.

chrome_2018-08-16_15-31-52


kirsty-unosquare (approved):
Nice tidy up. All good 👍

paulmatchett (approved):

PR #165, opened 2018-08-16
Feat/user profile style native

lee-cheung:
minor changes to log in
changes to header navigation and remaining/taken days

### DIFF
--- a/NativeApp/src/components/Common/HeaderDays/styles.js
+++ b/NativeApp/src/components/Common/HeaderDays/styles.js
@@ -11,9 +11,10 @@ const styles = StyleSheet.create({
     maxHeight: 70,
     borderColor: LIGHTGREY,
     borderTopWidth: 0.7,
+
   },
   holidayContainer: {
-    marginLeft: '5%',
+    marginLeft: 20,
     flex: 1,
     borderTopWidth: 1,
     borderColor: LIGHTGREY,
@@ -26,13 +27,8 @@ const styles = StyleSheet.create({
   h2: {
     color: GREY,
   },
-  h1: {
-    fontWeight: 'bold',
-  },
-  p: {
-    fontWeight: 'bold',
-    paddingBottom: 2,
-    paddingLeft: 2,
+  daysText: {
+    paddingLeft: 3,
   },
 });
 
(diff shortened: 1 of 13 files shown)


lee-cheung:
screenshot_1534497294
screenshot_1534497077

William-Unosquare:
Design-wise I think there are too many font sizes in the remaining/taken days info. I'd suggest maybe keeping the "days" part of the text the same size as the number?

Reducing this stuff helps designs look less chaotic.

NativeApp/src/components/Common/HeaderDays/styles.js
```diff
@@ -1,45 +1,37 @@
...
+    backgroundColor: WHITE,
+    paddingTop: 11,
+    paddingBottom: 11,
+    maxHeight: 70,
+    borderColor: LIGHTGREY,
+    borderTopWidth: 0.7,
   },
   holidayContainer: {
-    alignItems: 'center',
+    marginLeft: '5%',
```
kirsty-unosquare: Is there a reason for using a percentage for margin left? I'd rather if you used `justifyContent` or something similar to do with flex. Then add a number for marginLeft, e.g `marginLeft: 3`.

I'm not a big fan of using percentage for containers as it might look off on different devices. Especially devices with wider screens.

NativeApp/src/components/User/View/styles.js
```diff
@@ -34,15 +34,12 @@ const styles = StyleSheet.create({
...
-    paddingTop: 10,
     alignSelf: 'center',
-    flexDirection: 'row',
   },
   flatListView: {
     flex: 1,
   },
-  H3Bold: {
+  profileText: {
     fontWeight: 'bold',
```
kirsty-unosquare: Noticing a lot of bold font type being added to Texts. Maybe better add a bold type to the Texts common components?

kirsty-unosquare (changes requested):
Just a couple of comments. Mostly to do with keeping fonts consistent and maintainable. Noticed the bold font weight being added in a few places. Be good to move into the common components as a type.

NativeApp/src/components/Common/HeaderDays/styles.js
lee-cheung: i have changed the marginLeft to `marginLeft: 20`

NativeApp/src/components/User/View/styles.js
lee-cheung: Yes, i agree. I have added the bold type to the component.

kirsty-unosquare (approved):

PR #167, opened 2018-08-17
[TS-27] Add conditional checks to show approval buttons to holiday modal

William-Unosquare:
- Added `showAdminControls` prop to `HolidayModal`. Defaults to true as it's the most used use-case.
- Admin controls will now not show in holiday modal if the user is not an admin.
- Admin controls will now not show in holiday modal if an admin is viewing their own holidays.

### DIFF
--- a/web-app/src/components/HolidayModal/container.js
+++ b/web-app/src/components/HolidayModal/container.js
@@ -1,16 +1,25 @@
 import React, { Component } from 'react';
 import { PropTypes as PT } from 'prop-types';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
 import { isEmpty } from 'lodash';
 import { approveHoliday, rejectHoliday } from '../../services/holidayService';
 import swal from 'sweetalert2';
 import holidayStatus from '../../utilities/holidayStatus';
 import { Toast } from '../../utilities/Notifications';
+import { getUser } from '../../reducers';
 
-export default Wrapped =>
+const HolidayModalContainer = Wrapped =>
   class extends Component {
     static propTypes = {
       holiday: PT.object.isRequired,
       closeModal: PT.func.isRequired,
+      showAdminControls: PT.bool,
+      userDetails: PT.object.isRequired,
+    };
+
+    static defaultProps = {
+      showAdminControls: true,
     };
 
     constructor(props) {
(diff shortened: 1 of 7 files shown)


web-app/src/components/HolidayModal/container.js
```diff
@@ -1,16 +1,25 @@
...
   class extends Component {
     static propTypes = {
       holiday: PT.object.isRequired,
       closeModal: PT.func.isRequired,
+      showAdminControls: PT.bool,
+      userDetails: PT.object.isRequired,
+    };
+
+    static defaultProps = {
+      showAdminControls: true,
```
d1wilko: should the default here not be `false`? - I appreciate the default is unlikely to ever be hit but as a fallback I assume you don't want to show admin controls

web-app/src/components/HolidayModal/index.js
```diff
@@ -5,15 +5,26 @@ import { Modal, Button, Email } from '../../components/common';
...
   rejectHoliday,
+  userDetails,
+  showAdminControls,
 }) => {
   const { start, end, employee, eventStatus } = holiday;
-  const { forename, surname, email } = employee;
+  const { forename, surname, email, employeeId } = employee;
+  const isAdmin = userDetails.employeeRoleId === roles.ADMIN;
+
+  const shouldShowAdminControls = () => {
```
d1wilko, line 22: just a suggestion - I don't mind if you disregard - would this not be better as a pure function?

i.e. take in args as opposed to relying on external state?

d1wilko (changes requested):
looks good to me except - 1 issue and 1 suggestion

web-app/src/components/HolidayModal/index.js
William-Unosquare: I'm not sure what you mean, could you go into more detail?

web-app/src/components/HolidayModal/container.js
William-Unosquare: Fair point. I'll change this. I just didn't want to end up with a massive PR with lots of file changes.

web-app/src/components/HolidayModal/index.js
d1wilko: I'm going to approve, but what I mean is, instead of `const shouldShowAdminControls = () => {`

go for something like `const shouldShowAdminControls = (isAdmin, userDetails, employeeId, showAdminControls) => {`

this would mean that the function is pure and doesn't rely on external state - respecting the whole javascript as a functional language paradigm.

d1wilko (approved):
suggestion added - but approving

PR #168, opened 2018-08-17
Add fonts to native app

kirsty-unosquare:
I've added in custom fonts using expo's sdk Font ([Expo Font](https://docs.expo.io/versions/v29.0.0/sdk/font)). Had to configure `AuthContext` to ensure the app only loads when fonts have been loaded in. Throws an error if Font is used before loaded.

I decided to do this top level ( booting process ) instead of checking all the Text components. My only concern is what happens if fonts are never loaded. I couldn't find a way to handle this. Any one else know?


William-Unosquare:
In what cases would the font never load? Are they packaged into the app or do the fonts come from the network?

kirsty-unosquare:
Not coming from web, packaged into app under assets folder. Think it should be okay, just paranoid that it's not being handle. 😅

William-Unosquare:
Yeah if it's not coming from the network I'm pretty sure that the checks you have in place here are enough to protect from any issues with them not being loaded in.

William-Unosquare (approved):

lee-cheung (approved):

d1wilko (approved):
good clean code

PR #169, opened 2018-08-20
Fix and refactor to legend / filter component

William-Unosquare:
The code for this PR is a bit complicated. Lots of data being worked with. Any questions just shoot to myself or @paul-matchett. The main chunk of logic to be concerned about is in `Dashboard/container.js`

The PR is a refactor to how we were filtering the events in the calendar, which hopefully future-proofs this a bit better for any upcoming changes to the dashboard API. It also contains bug fixes for issues that the previous solution had.


web-app/src/components/Legend/container.js
```diff
@@ -12,130 +12,91 @@ const LegendContainer = Wrapped =>
...
+    };
 
-      if (employees.length === prevEmployees.length && employees.length === 0) {
-        this.createEmployeesList(takenHolidays);
+    componentDidUpdate = (_, prevState) => {
+      const { employeeList } = this.state;
+      const prevEmployees = prevState.employeeList;
+      if (
+        employeeList.length === prevEmployees.length &&
+        employeeList.length === 0
```
d1wilko: a bit confused by this logic - this will only fire if...
`employeeList.length === 0` && `employeeList.length === prevEmployees.length`
which means...
`prevEmployees.length === 0`
is there a clearer way to write this?

web-app/src/components/Legend/container.js
```diff
@@ -12,130 +12,91 @@ const LegendContainer = Wrapped =>
...
-      employees = this.removeDuplicateEmployees(employees);
-      employees.unshift({ value: -1, displayValue: 'All' });
-      this.setState({ employees: this.sortEmployeeList(employees) });
     };
 
-    setKeyActiveState = eventStatusId => {
-      let updatedList = [...this.state.eventsKeys];
+    storeEmployeeListToState = () => {
+      const { takenHolidays } = this.props;
+      const employeeList = takenHolidays
```
d1wilko: from line 67 to line 77 this gets really hard to read - I'm confident this could be made simpler

web-app/src/components/Legend/container.js
```diff
@@ -12,130 +12,91 @@ const LegendContainer = Wrapped =>
...
-      this.setState({
-        selectedEventStatusID: eventStatusId,
-      });
-      this.props.updateCalendarEvents(eventStatusId);
+      this.setState({ [updatedProp]: updatedList });
+      this.props.updateCalendarEvents(eventId, event);
     };
 
-    handleFormStatus(name, value) {
+    onFilterUserChange(_, value) {
```
d1wilko: is the underscore needed here?

web-app/src/components/Legend/container.js
```diff
@@ -144,20 +105,21 @@ const LegendContainer = Wrapped =>
...
 
     render() {
       return (
         <Wrapped
-          employees={this.state.employees}
           selectedEmployee={this.state.selectedEmployee}
+          employeeList={this.state.employeeList}
           formStatus={(name, value, formIsValid) =>
-            this.handleFormStatus(name, value, formIsValid)
+            this.onFilterUserChange(name, value, formIsValid)
```
d1wilko: is `name` of `formIsValid` needed here? doesn't seem to be used

d1wilko (changes requested):
theres a lot of code to go through here - I'd recommend getting @bgribz to have a look through aswell

web-app/src/components/Legend/container.js
paulmatchett: The name param is not required for this function even though it is being passed in. saw 
 underscore being used as a param for componentDidUpdate in pages -> dashboard -> container and thought i'd reuse. Is this bad practice?

web-app/src/components/Legend/container.js
paulmatchett: No it's not. This will be removed.

web-app/src/components/Legend/container.js
William-Unosquare: @paul-matchett If `name` is never being used in the function any time it's being called there is no point of it being a parameter. The usage in componentDidMount is because we don't want to use `prevProps` but we need access to the second parameter `prevState`, however, there are obvious use cases where prevProps is required. 

TypeScript sees this as expected behaviour (according to VSCode errors) and it's practice in other languages.

William-Unosquare:
The real problem around that underscore was the fact that an entire form component was being used for one field. This didn't need validation or anything other than styling that the form component gives us so I've refactored as well. This PR is getting quite large now so hopefully, that's all the work needing done to it.

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +12,110 @@ const LegendContainer = Wrapped =>
...
+        .filter(hol => hol.employee)
+        .map(hol => hol.employee)
+        .filter((employee, pos, arr) => {
+          return (
+            arr
+              .map(mapObj => mapObj.employeeId)
+              .indexOf(employee.employeeId) === pos
+          );
+        });
+      this.setState({ employeeList });
```
bgribz: I don't understand why this is being kept in state if it is never being modified.

Renaming this function to something like `getEmployeeList` or similar and calling it in the render method would make more sense.

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +12,110 @@ const LegendContainer = Wrapped =>
...
     };
 
-    createEmployeeObject = hol => {
-      let { employeeId, forename, surname } = hol.employee;
-      return {
-        value: employeeId,
-        displayValue: `${forename} ${surname}`,
-      };
-    };
+    storeLegendKeysToState = () => {
```
bgribz: I don't understand why this is being kept in state if it is never being modified.

web-app/src/components/Legend/container.js
bgribz: Yeah I agree with Dan here - if I had to guess I would say it's getting a list of employees from the list of holidays?

If so, `const employeeList = takenHolidays.map(x => x.employee)` would do the job. If there might be issues with undefined values we can make use of lodash:
```  
const employeeList = _
  .chain(takenHolidays)
  .map('employee')
  .compact()
  .value();
```

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +12,110 @@ const LegendContainer = Wrapped =>
...
+          );
+        });
+      this.setState({ employeeList });
+    };
+
+    setLegendKeyActiveState = (eventId, event) => {
+      let updatedList;
+      let updatedProp;
+      if (event === 'status') {
+        updatedList = [...this.state.statusList];
```
bgribz: Rather than maintaining two separate lists for events, I think we should keep it as one and filter the data being passed into our Wrapped component. It would make this function a lot cleaner/clearer.

Also, it might be worth looking into whether or not we need this complexity around these keyObjects and lists. Would it not be easier to maintain a single array of active event filter IDs or something? The fact that the event statuses/types are a constant piece of data, I kinda feel like we could be smarter about how we generate/manage the components.

Also also, it's 22:34 and this is just me thinking out loud.

web-app/src/components/common_styled/Select.js
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components';
```
bgribz, line 1: What's the reason behind a `common` and `common_styled` folder split?

bgribz (changes requested):

web-app/src/components/Legend/container.js
paulmatchett: Never thought of using lodash to be honest, but yeah, i've updated it and it looks so much cleaner.

```javascript
const employeeList = _.chain(takenHolidays)
        .map('employee')
        .compact()
        .uniqBy('employeeId')
        .value();
```

Thanks @d1wilko and @bgribz

web-app/src/components/Legend/container.js
paulmatchett: Your right. The function is now being called from the render method and employeeList has been removed from state

web-app/src/components/common_styled/Select.js
William-Unosquare: Personal preference really. `common_styled` is to keep components that deal with style only in their own place and easy to find.

bgribz: Cannae say I'm a fan, but as you said personal preference 👍🏻

William-Unosquare:
I'm really sorry for how messy this PR has got. The scope of the refactoring needed here has been pretty large. It's going to need another scan through by someone after this commit which changes quite a bit:
https://github.com/UnosquareBelfast/AdminCore/pull/169/commits/160590a6113ca79ab998c6971254f36dd3ef3125

d1wilko:
Going to look at this now, but, this is a huge code review and a clear example of why we should never deliver code with the intention of refactoring later

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +13,89 @@ const LegendContainer = Wrapped =>
...
-            obj => obj.value === o.value && obj.displayValue === o.displayValue
-          )
-        ) {
-          unique.push(o);
-        }
-        return unique;
-      }, []);
-    };
+      for (let i = 0; i < legendKeys.length; i++) {
+        const legendKey = legendKeys[i];
```
d1wilko: this would be easier to read as a foreach
`legendKeys.forEach((legendKey, index) => {...`

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +13,89 @@ const LegendContainer = Wrapped =>
...
     };
 
-    setKeyActiveState = eventStatusId => {
-      let updatedList = [...this.state.eventsKeys];
-      for (let key of updatedList) {
-        if (key.eventStatusId === eventStatusId) {
+    toggleKeyActive = keyId => {
+      const legendKeys = [...this.state.legendKeys];
+      legendKeys.map(key => {
+        if (key.id == keyId) {
```
d1wilko: best to use `===` here - type conversion can lead to unexpected issues

bgribz:
I can't see any glaring issues with this, but going forward we should be conscious about the size of these PRs.

bgribz (approved):

web-app/src/components/Legend/container.js
```diff
@@ -12,152 +13,89 @@ const LegendContainer = Wrapped =>
...
         () => {
           this.props.updateEmployee(this.state.selectedEmployee);
         }
       );
-    }
+    };
+
+    getEmployeeState = () => {
+      const { takenHolidays } = this.props;
+      return _.chain(takenHolidays)
```
d1wilko: noice

William-Unosquare:
@d1wilko and @bgribz  Noted, but also a reminder that this is quickly changing on the front and back end so sometimes these refactors are unavoidable if a change somewhere causes something to break or creates a scenario where a refactor helps maintain the code. This was one of those cases.

web-app/src/pages/Dashboard/container.js
```diff
@@ -11,90 +11,88 @@ const DashboardContainer = Wrapped =>
...
 
     componentDidMount() {
       this.props.fetchEvents();
     }
 
-    componentDidUpdate = (_, prevState) => {
-      if (prevState.selectedEmployeeId !== this.state.selectedEmployeeId) {
-        this.updateFilterEvents();
+    componentDidUpdate = prevProps => {
+      if (prevProps.takenEvents != this.props.takenEvents) {
```
d1wilko: !==

web-app/src/pages/Dashboard/container.js
```diff
@@ -11,90 +11,88 @@ const DashboardContainer = Wrapped =>
...
+            takenEvents={this.props.takenEvents}
+            events={this.state.filteredEvents}
+            updateTakenEvents={this.props.fetchEvents}
             isEventBeingUpdated={this.props.isEventBeingUpdated}
-            onUpdateEvents={this.onFilterEvents}
-            onUpdateEmployee={this.onFilterEmployee}
+            onUpdateEvents={activeEventIds =>
+              this.setActiveEvents(activeEventIds)
+            }
+            onUpdateEmployee={({ employeeId }) =>
```
d1wilko: I might have confused myself here, but why is `{ employeeId }` destructured? can this not just be `onUpdateEmployee={employeeId =>`? apologies if I've got this wrong

d1wilko (changes requested):
some changes requested, but definitely in better shape

web-app/src/pages/Dashboard/container.js
bgribz: I think this gets passed an employee object

web-app/src/components/Legend/container.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import { PropTypes as PT } from 'prop-types';
-import holidayStatus from '../../utilities/holidayStatus';
+import _ from 'lodash';
```
William-Unosquare: @paul-matchett You're importing the entirety of lodash here. Import only what you need. This may be because of the chain method you're using so give this a read:
https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba

look into lodash/fp/flow

d1wilko (approved):
looks good to me - I personally like the chain stuff but no issue with it changing to flow

PR #173, opened 2018-08-22
Fix/days notice warning

paulmatchett:
Notice warning Message not displaying correctly when required. Fixes for this included:

- DaysDuration needed to be passed down as a prop to bookingModal. This value is not required initially, but was given a default value of 1.
- Message not displaying when re selecting annual holiday. eventTypeId was being checked against a string instead of an number.


web-app/src/components/BookingModal/BookingModalForm/container.js
```diff
@@ -67,7 +72,10 @@ const Container = Wrapped =>
         }
       } else if (name === 'isHalfday' && formData.isHalfday) {
         formData.end = formData.start;
-        formData.eventTypeId = 1;
+        const annualLeave = 1;
+        formData.eventTypeId = annualLeave;
```
William-Unosquare: Would it not be better to use `formData.eventTypeId = eventTypes.ANNUAL_LEAVE` here?

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -42,13 +42,14 @@ const BookingModalForm = props => {
 
   const composeErrorMessage = () => {
     const { eventTypeId, start } = formData;
-    if (isEventBeingUpdated || eventTypeId !== 1) {
+    const annualLeave = 1;
+    if (isEventBeingUpdated || eventTypeId !== annualLeave) {
```
BrownMark: Same as William's comment above, we have constants for these

web-app/src/components/BookingModal/BookingModalForm/container.js
paulmatchett: @William-Unosquare - Done. Also changed HOLIDAY to ANNUAL_LEAVE in eventTypes and updated files where neccessary. See updated commits below

web-app/src/components/BookingModal/BookingModalForm/index.js
paulmatchett: @BrownMark  - rookie mistake. this has been updated

William-Unosquare (approved):

PR #174, opened 2018-08-23
Update/validate event selection

paulmatchett:
On the dashboard calendar When a user selects dates to request, validation will now check:

- if the date range falls in the past
- if the selected date range overlaps over events already booked off by that user

Please note that moment-range has been added to package.json, so 'npm i' command will need to executed before starting up the application


web-app/src/components/BookingCalendar/container.js
```diff
@@ -36,21 +40,65 @@ const BookingCalendarContainer = Wrapped =>
...
+
+    selectedDatesOverlapExisting = (start, end) => {
+      const overlappingEvents = this.props.takenHolidays.filter(hol => {
+        if (hol.employee && hol.employee.employeeId === this.props.employeeId) {
+          var selectedDateRange = moment.range(
+            moment(start),
+            moment(end).endOf('day'),
+          );
+          var existingEvent = moment.range(moment(hol.start), moment(hol.end));
+          if (selectedDateRange.overlaps(existingEvent)) {
```
William-Unosquare, line 58: Love this - nice and clean.

web-app/src/components/BookingCalendar/container.js
```diff
@@ -36,21 +40,65 @@ const BookingCalendarContainer = Wrapped =>
...
-    onSelectSlot = ({ start, end }) => {
+    checkIfPastDatesSelected = start => {
       const today = new moment();
-      if (moment(start).isAfter(today.add(-1, 'days'))) {
+      return moment(start).isBefore(today);
+    };
+
+    selectedDatesOverlapExisting = (start, end) => {
+      const overlappingEvents = this.props.takenHolidays.filter(hol => {
+        if (hol.employee && hol.employee.employeeId === this.props.employeeId) {
```
William-Unosquare: Destructure these props

web-app/src/components/BookingCalendar/container.js
```diff
@@ -36,21 +40,65 @@ const BookingCalendarContainer = Wrapped =>
...
+            moment(start),
+            moment(end).endOf('day'),
+          );
+          var existingEvent = moment.range(moment(hol.start), moment(hol.end));
+          if (selectedDateRange.overlaps(existingEvent)) {
+            return true;
+          }
+        }
+      });
+      return overlappingEvents.length > 0 ? true : false;
```
William-Unosquare: I don't think you need the `?  true : false` here. The comparison operator will return these anyway.

web-app/src/components/BookingCalendar/container.js
```diff
@@ -36,21 +40,65 @@ const BookingCalendarContainer = Wrapped =>
...
+        if (datesOverlapExisting) {
+          return [
+            false,
+            'Your are trying to request dates that have already been set',
+          ];
+        } else {
+          return [true, 'dates approved'];
+        }
+      }
+    };
```
William-Unosquare: Not sure we need the complexity of using an array as a return value here.

It could just return a error message if there's an error, and nothing if it's successful. Then below use `title: error || 'Dates approved'`,

William-Unosquare (changes requested):

web-app/src/components/BookingCalendar/container.js
paulmatchett: I didn't destructure them as employee might not exists in the hol object. hol.employee need to exist first, which is the first check in the if statement.

web-app/src/components/BookingCalendar/container.js
paulmatchett: Updated to return a string. If string return equals 'Dates approved' it opens modal, otherwise displays error message

web-app/src/components/BookingCalendar/container.js
William-Unosquare: That shouldn't matter. You can still destructure, it's just if `hol.employee` isn't there then the destructured `employee` would be given the value of `undefined`. I had to double check this but:

code_2018-08-23_13-44-50

So you can just do
`if (employee && employee.employeeId === employeeId)`

web-app/src/components/BookingCalendar/container.js
William-Unosquare: Not a huge fan of this approach either. It feels brittle to have logic tied to what we're rendering. But I'll not get hung up in it.

web-app/src/components/BookingCalendar/container.js
paulmatchett: ok, updated and pushed up. Personality, I think its not needed and adds another line of code

web-app/src/components/BookingCalendar/container.js
```diff
@@ -47,7 +47,8 @@ const BookingCalendarContainer = Wrapped =>
 
     selectedDatesOverlapExisting = (start, end) => {
       const overlappingEvents = this.props.takenHolidays.filter(hol => {
-        if (hol.employee && hol.employee.employeeId === this.props.employeeId) {
+        const { employee } = hol;
+        if (employee && employee.employeeId === this.props.employeeId) {
```
William-Unosquare: You still haven't destructured the props haha..

`this.props.employeeId` and `this.props.takenHolidays`

paulmatchett: Just for you.... see commit below

William-Unosquare: Uck, thanks

William-Unosquare (approved):

PR #176, opened 2018-08-23
Implement new dashboard call

William-Unosquare:
- Dashboard now only shows events for the logged in user for the displayed month.
- When the new call is responding it is intercepted. At this point, the dates are changed to moment objects and we tack on information about the currently logged in user to each event for seamless integration with the current calendar logic.
- Cleaned up some stuff that isn't used any more.

I haven't yet made the request fire again when the month is changed to keep this PR nice and small.

### DIFF
--- a/web-app/src/actions/dashboard.js
+++ b/web-app/src/actions/dashboard.js
@@ -1,5 +1,4 @@
 import * as actionTypes from '../actionTypes';
-import { getHolidays } from '../services/holidayService';
 import { getUsersEvents } from '../services/dashboardService';
 
 import {
--- a/web-app/src/actions/dashboard.js
+++ b/web-app/src/actions/dashboard.js
@@ -1,5 +1,7 @@
 import * as actionTypes from '../actionTypes';
-import { getAllHolidays, getHolidays } from '../services/holidayService';
+import { getHolidays } from '../services/holidayService';
+import { getUsersEvents } from '../services/dashboardService';
+
 import {
   formatEventsForCalendar,
   getEventDuration,
@@ -53,19 +55,8 @@ export const setError = error => {
 
 // Thunks
 
-export const fetchEvents = () => dispatch => {
-  getAllHolidays()
-    .then(({ data }) => {
-      const formattedEvents = formatEventsForCalendar(data);
-      dispatch(setCalendarEvents(formattedEvents));
-    })
-    .catch(error => {
-      dispatch(setError(error));
-    });
-};
-
-export const fetchEventsByUserId = userId => dispatch => {
-  getHolidays(userId)
+export const fetchEvents = date => dispatch => {
+  getUsersEvents(date)
     .then(({ data }) => {
       const formattedEvents = formatEventsForCalendar(data);
       dispatch(setCalendarEvents(formattedEvents));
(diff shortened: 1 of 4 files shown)


web-app/src/actions/dashboard.js
```diff
@@ -1,5 +1,7 @@
 import * as actionTypes from '../actionTypes';
-import { getAllHolidays, getHolidays } from '../services/holidayService';
+import { getHolidays } from '../services/holidayService';
```
paulmatchett: getHolidays is not being used? Will this come into play when events for multiple people are required?

web-app/src/pages/Dashboard/container.js
```diff
@@ -2,30 +2,32 @@ import React from 'react';
...
       fetchEvents: PT.func.isRequired,
-      fetchEventsByUserId: PT.func.isRequired,
       takenEvents: PT.array,
       isEventBeingUpdated: PT.bool,
     };
 
     constructor(props) {
       super(props);
       this.state = {
+        calendarDate: new moment(),
```
paulmatchett, line 21: Could we not format the date here? ie, new moment.format('YYYY-MM-DD')

web-app/src/pages/Dashboard/container.js
```diff
@@ -2,30 +2,32 @@ import React from 'react';
...
         filteredEvents: [],
         activeEventIds: [],
         activeEmployee: -1,
       };
     }
 
     componentDidMount() {
-      this.props.fetchEvents();
+      const { calendarDate } = this.state;
```
paulmatchett: Could we not call fetchEvents method instead. save duplicating code?

web-app/src/utilities/AxiosInstance.js
```diff
@@ -37,6 +38,24 @@ instance.interceptors.response.use(function(response) {
     };
   }
 
+  // Append employee to each event (we know its the logged in user) and convert
+  // dates to moment objects
+  if (response.config.url.includes(`${baseURL}/dashboard/getEmployeeEvents`)) {
+    const events = [...response.data.events];
+    const employee = store.getState().USER;
+    for (const index in events) {
```
d1wilko: small suggestion but I would use `property` as opposed to `index` here - as I believe events is an object as opposed to an array? this will make the code a bit more self documenting and understandable.

If events is an array as opposed to an object you should use a for OF loop

d1wilko (commented):
one small change

web-app/src/pages/Dashboard/container.js
William-Unosquare: It's best left like this, I think. Keep in mind the next PR will be changing this value by calling other moment methods on this moment instance.

web-app/src/pages/Dashboard/container.js
William-Unosquare: Ayup. Good spot.

web-app/src/utilities/AxiosInstance.js
William-Unosquare: It's an array of objects. Changed.

web-app/src/actions/dashboard.js
William-Unosquare: Another good spot 👍 Just leftovers from the other stuff I removed in this file.

paulmatchett (approved):

PR #177, opened 2018-08-24
Update/adding app icons

paulmatchett:
Installed FaviconsWebpackPlugin plugin and added to webpack.config file to dynamically create a manifest file and also flavicon, ios, android icons and splash screens based on one unosquare logo, as well as all the necessary meta tags to go with.


William-Unosquare (approved):
Looks good to me. But get someone else to check this that is more familiar with this stuff.

ghost (approved):
I like this.

BrownMark (approved):
Looks good, adds some time to compile but it's worth it.

PR #180, opened 2018-08-24
updated styles to make more events link visible and styled more links

paulmatchett:
Small update to fix the ui for more than three events on same calendar day

capture

image

### DIFF
--- a/web-app/src/components/common/Event/styled.js
+++ b/web-app/src/components/common/Event/styled.js
@@ -6,7 +6,8 @@ export const Container = styled.div`
   white-space: nowrap;
   overflow: hidden;
   cursor: pointer;
-  padding: 7px 5px;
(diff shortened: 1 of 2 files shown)


William-Unosquare:
The event boxes are looking very thin again. Was there no other way to implement this other than reducing the height of the event boxes?

paulmatchett:
The plugin seems to adjust the height of the table rows based on their styling of the events. I tried keeping them the previous size but increasing the height of the calendar (therefore increasing the row height), but for some reason the more links don't show at all.

William-Unosquare:
Yeah, I found discussion about that here:
https://github.com/intljusticemission/react-big-calendar/issues/679

That's a real shame.

William-Unosquare (approved):

PR #181, opened 2018-08-24
Changing month now requests data for that month

William-Unosquare:
When the calendar is moved back or forth in months a new request is fired to populate the events for that month.

Again, keeping these PR's small so still to do as far as I'm concerned:
- Append new months to existing months instead of overwrite (this could be tough to do)
- Create a full-page loading state with redux to trigger while these requests are being made.

### DIFF
--- a/web-app/src/components/BookingCalendar/container.js
+++ b/web-app/src/components/BookingCalendar/container.js
@@ -127,19 +127,15 @@ const BookingCalendarContainer = Wrapped =>
       }
     };
 
-    onNavigate = date => {
-      const newDate = new moment(date);
-      this.props.onNavigate(newDate);
-    };
-
     render() {
+      const { employeeId, events, onNavigate } = this.props;
       return (
-        this.props.employeeId &&
-        this.props.events && (
+        employeeId &&
+        events && (
           <Wrapped
             onSelectSlot={this.onSelectSlot}
             onSelectEvent={this.onSelectEvent}
-            onNavigate={this.onNavigate}
+            onNavigate={onNavigate}
             events={this.props.events}
           />
         )
--- a/web-app/src/components/BookingCalendar/container.js
+++ b/web-app/src/components/BookingCalendar/container.js
@@ -24,6 +24,7 @@ const BookingCalendarContainer = Wrapped =>
       updateEventDuration: PT.func,
       setEventBeingUpdated: PT.func,
       toggleBookingModal: PT.func,
+      onNavigate: PT.func,
     };
 
     constructor(props) {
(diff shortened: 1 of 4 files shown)


web-app/src/components/BookingCalendar/index.jsx
```diff
@@ -10,7 +10,13 @@ moment.locale('en-gb');
 Calendar.momentLocalizer(moment);
 
 export const BookingCalendar = props => {
-  const { onSelectSlot, onSelectEvent, events } = props;
+  const { onSelectSlot, onSelectEvent, events, onNavigate } = props;
+
+  const updateOnNavigate = date => {
```
paulmatchett: Is updateOnNavigate  method required? Its already on the bookingCalendar container.js and being passed down as a prop.

William-Unosquare: Yea you're right. I think I made that to avoid this:
`onNavigate={date => onNavigate(new moment(date))}` which is kinda ugly.

But I've updated it now to transform the date into a moment further down the line instead.

paulmatchett (approved):

web-app/src/components/BookingCalendar/container.js
```diff
@@ -126,13 +127,19 @@ const BookingCalendarContainer = Wrapped =>
       }
     };
 
+    onNavigate = date => {
+      const newDate = new moment(date);
+      this.props.onNavigate(newDate);
```
ghost: Naming is a bit weird here or I'm being silly. You have a function called onNavigate, inside you get a new formatted date and pass it to a new variable newDate. Then invoke the function onNavigate from the props passing in the newDate variable? 

I'd be renaming either the prop or the function name personally.

web-app/src/pages/Dashboard/container.js
```diff
@@ -79,9 +79,19 @@ const DashboardContainer = Wrapped =>
       this.setState({ activeEmployee: employeeId }, this.filterCalenderEvents);
     };
 
+    handleCalendarNavigate = date => {
+      const newDate = new moment(date);
+      this.setState(
+        {
+          calendarDate: newDate.startOf('month').format('YYYY-MM-DD'),
```
ghost, line 86: Is calendarDate not already set to be that `new moment().startOf('month').format('YYYY-MM-DD')` within the constructor so you don't need to do it within handleCalendarNavigate. Just assuming here, I could be wrong.

William-Unosquare: `calenderDate` is initially set to `new moment().startOf('month').format('YYYY-MM-DD')`, but here we're updating it using `date` which comes from the calendar component, `date` is a javascript object rather than a moment object so we transform it into a moment object (`newDate`) which means we need to do the date modification and formatting again.

In short, we're setting state from a completely different object, so we need to transform it again.

web-app/src/components/BookingCalendar/container.js
William-Unosquare: This actually isn't even needed at all after some changes I made later on. :)

web-app/src/pages/Dashboard/container.js
ghost: Oops didn't see the start of it! Sounds good!

ghost (approved):

PR #185, opened 2018-08-27
updating 1st query with new logic to get most relevant events

eoinmcafee:


BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/DashboardRepository.java
```diff
@@ -10,13 +10,25 @@
 
 
 public interface DashboardRepository extends JpaRepository<Event, Integer> {
-    @Query(
-            value = "SELECT * FROM Event e WHERE (e.start_Date >= :startDate AND e.end_Date <= :endDate) AND e.employee_id = :employeeId",
+    @Query(value = "SELECT * FROM Event e INNER JOIN Contract c on :employeeId = c.employee_id " +
```
BrownMark: This join is a little confusing, i'd join 'e' to 'c' here and add the employee check in the where.

BrownMark (changes requested):

BrownMark (approved):

PR #188, opened 2018-08-27
quick fix on selecting 4 days not displaying duration warning

paulmatchett:
Quick fix where selecting 4 days right after current day, not displaying warning


William-Unosquare (approved):
👌🏻

PR #190, opened 2018-08-27
Clean up dashboard events

William-Unosquare:
This is just a lot of clean up. It's an ugly PR to review but shouldn't be too bad if you view each commit rather than all the changes at once.

Lots of this PR is just renaming stuff better. Should put us in a better place to add another layer of complexity if we need to.


ghost (approved):
Looks good!

paulmatchett (approved):

PR #191, opened 2018-08-28
Added a toggle button to switch between a user and team events

paulmatchett:
- Created a generic toggle button
- Created an eventView utility to hold dashboard switch view toggle button icons and text
- added the toggle button the the dashboard
- have the button toggle a new state property called eventView 

**The toggle button is only updating the dashboards state eventView property that is being used to identify the buttons active styling. Events still need to toggle based on the active state.** 

An additional PR will be made to display the buttons on all device sizes. Currently only works on desktop

image

### DIFF
--- a/web-app/src/components/common/ToggleButton/index.js
+++ b/web-app/src/components/common/ToggleButton/index.js
@@ -1,0 +1,33 @@
+import React from 'react';
+import { PropTypes as PT } from 'prop-types';
+import { ButtonToggle } from './styled';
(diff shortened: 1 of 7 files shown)


William-Unosquare (approved):
Like we talked about on Slack I'd rather the state for the eventView was handled within Redux to keep everything Dashboard consistently controlled there. But I'm going to approve this so you can continue work on the responsive stuff.

PR #192, opened 2018-08-28
Fix for holidayId/eventId

paulbrownlee:
This pull request changes the holidayId to eventId to keep consistency .


web-app/src/utilities/AxiosInstance.js
```diff
@@ -2,8 +2,8 @@ import axios from 'axios';
 import moment from 'moment';
 import store from '../store';
 
-const baseURL = process.env.DOMAIN;
-
+//const baseURL = process.env.DOMAIN;
+const baseURL = 'http://localhost:8081';
```
BrownMark: You'll need to undo this. In future create a .env file with DOMAIN set up.

paulbrownlee: Just pushed up the fixed AxiosInstance

BrownMark:
What's the reason behind this PR? I would've though from the holiday controller we'd be specifying holiday IDs?

Is it to help out the FE so that they can use eventID everywhere?

paulbrownlee:
William requested it in order for the front-end to keep the eventID consistent.

William-Unosquare:
@BrownMark Yea this helps the front-end for that reason. I'd mentioned it in the internal chat on Slack a few days back. If this makes the back-end horrible though, we can still work with holidayId

eoinmcafee (approved):

eoinmcafee (approved):

eoinmcafee (approved):

PR #195, opened 2018-08-28
Fetch employee events for each month.

kirsty-unosquare:
Updated fetching event call for calendar. Also updated some terms to match new event call. Let me know if you see anywhere that needs updated. Very similar to web app.

Had to use the holidays call for profile screen since I need to retrieve all events. Currently the profile screen is broken but will fix on another branch.

### DIFF
--- a/NativeApp/src/components/User/View/index.js
+++ b/NativeApp/src/components/User/View/index.js
@@ -14,7 +14,7 @@ import getDuration from '../../../utilities/dates';
 
 
 const UserView = (props) => {
-  const { takenHolidays, remainingHolidays, employee } = props;
+  const { events, remainingHolidays, employee } = props;
 
   return (
     <SafeAreaView style={{ flex: 1, backgroundColor: WHITE }}>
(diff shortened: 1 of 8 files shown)


William-Unosquare (approved):
All looks good to me. :)

d1wilko (approved):
nothing to call out - nice code

NativeApp/src/components/User/View/index.js
```diff
@@ -27,14 +27,14 @@ const UserView = (props) => {
...
-          takenHolidays={takenHolidays}
+          events={events}
           remainingHolidays={remainingHolidays}
         />
         <ScrollView>
           <View style={styles.flatListView}>
             <FlatList
-              keyExtractor={item => item.holidayId.toString()}
-              data={takenHolidays}
+              keyExtractor={(item, index) => index.toString()}
```
paulmatchett: Can you not use eventId instead of index?

paulmatchett (approved):
Nice and neat. All looks good to me

NativeApp/src/components/User/View/index.js
kirsty-unosquare: Actually need to convert this back  to holiday id. The user screen still uses the 
 Get `/holidays/findByEmployeeId/{employeeId}` endpoint. I don't think eventId is listed in the response for this request.

Reverting back to holidayId, good spot!

PR #196, opened 2018-08-28
Tidyup: Dashboard event logic now goes through a pipe instead of funcs calling other funcs

William-Unosquare:
This tidies up the logic in place to transform the data coming from the server into something we can use with the calendar. Instead of functions calling other functions like spaghetti, it now follows the pipeline defined in the `transformEvents` function.

This tidy up is because the file is due to get more complicated.

### DIFF
--- a/web-app/src/utilities/mandatoryEvents.js
+++ b/web-app/src/utilities/mandatoryEvents.js
@@ -1,0 +1,35 @@
+import moment from 'moment';
+const currYear = new Date().getFullYear();
+
+// Add mandatory events here. The only fields that need change are title and the dates.
+const mandatoryEvents = [
+  {
+    title: 'Christmas Day',
+    mandatoryDate: `${currYear}-12-25`,
+  },
+  {
+    title: 'New Years Day',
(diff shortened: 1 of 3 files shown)


web-app/src/utilities/mandatoryEvents.js
```diff
@@ -0,0 +1,35 @@
...
+    title: event.title,
+    start: new moment([event.mandatoryDate], 'YYYY-MM-DD'),
+    end: new moment([event.mandatoryDate], 'YYYY-MM-DD'),
+
+    //Add the event boilerplate
+    eventId: -1,
+    allDay: true,
+    halfDay: false,
+    employee: null,
+    eventStatus: { eventStatusId: 4, description: 'Mandatory' },
```
paulmatchett: We should probably use the eventTypes const values instead of 4 and 1

paulmatchett (changes requested):
Great tidy up. Just left one comment regarding using eventTypes values instead of number.

paulmatchett (approved):

PR #197, opened 2018-08-29
Update/add calendar team view

paulmatchett:
Added team view to calendar and integrated it into redux


web-app/src/actions/dashboard.js
```diff
@@ -53,18 +54,39 @@ export const setError = error => {
   };
 };
 
+export const updateEventView = eventView => {
```
William-Unosquare: Can we rename this to setEventView? It's what you've called the actionType SET_EVENT_VIEW so keep it consistent here. I also feel like "set" better describes what this is doing.

William-Unosquare (changes requested):

William-Unosquare (approved):

PR #198, opened 2018-08-29
[TS-31] Ability to reject holidays

mccleaveryan:
Gives admin the ability to reject a holiday.


eoinmcafee:
Hey @mccleaveryan, 

There's another requirement to the story. The ability to add a reason why the request was cancelled. I'd expect a new field in the events table to store a comment and this being propagated through the stack. 

Cheers,
Eoin

mccleaveryan:
Apologies for missing the requirement. That'll be added in.

Cheers,
Ryan

BackEnd/src/main/java/com/unosquare/admin_core/back_end/enums/EventStatuses.java
```diff
@@ -8,7 +8,8 @@
     AWAITING_APPROVAL(1, "Awaiting approval"), //default clientStatus for new holiday requests
     APPROVED(2, "Approved"),
     DENIED(3, "Denied"),
-    CANCELLED(4,"Cancelled");
+    CANCELLED(4,"Cancelled"),
+    REJECTED(5,"Rejected");
```
BrownMark: What's the difference between denied and rejected?

BackEnd/src/main/java/com/unosquare/admin_core/back_end/service/EventService.java
```diff
@@ -99,6 +99,15 @@ public void cancelEvent(int eventId){
         }
     }
 
+    public void rejectEvent(int eventId){
+        Optional<Event> retrievedEvent = eventRepository.findById(eventId);
+        if (retrievedEvent.isPresent()) {
+            Event event = retrievedEvent.get();
+            event.setEventStatus(new EventStatus(EventStatuses.REJECTED.getEventStatusId()));
```
BrownMark: I don't think we need rejected as a event status, we can make use of denied?

BackEnd/src/main/java/com/unosquare/admin_core/back_end/service/EventService.java
```diff
@@ -99,6 +99,15 @@ public void cancelEvent(int eventId){
         }
     }
 
+    public void rejectEvent(int eventId){
```
BrownMark: surely you're building an EventDto from the RejectHolidayViewModel, would you not pass that in and map it down to the Event?

BrownMark (changes requested):
I'm not sure why the package.lock files changed

BackEnd/src/main/java/com/unosquare/admin_core/back_end/enums/EventStatuses.java
mccleaveryan: Makes sense to use Denied instead of a new status. I'll amend that. Thanks.

BackEnd/src/main/java/com/unosquare/admin_core/back_end/service/EventService.java
mccleaveryan: No problem.

BackEnd/src/main/java/com/unosquare/admin_core/back_end/configuration/mappings/domain/EventMappings.java
```diff
@@ -17,6 +17,7 @@ protected void configure() {
                 map().setEventType(new EventType(source.getEventTypeId()));
                 map().setStartDate(source.getStartDate());
                 map().setEndDate(source.getEndDate());
+                map().setDeniedMessage(source.getDeniedMessage());
```
eoinmcafee: I don't think we should couple the message to 'denied'. something like comment or message. That way it's property isn't just locked down to one event activity.

eoinmcafee:
@mccleaveryan is this the correct pull request? I pulled down the code and it's not compiling. It feels like the code is out of date?

PR #199, opened 2018-08-29
[TS-37] Append Events & Stop duplicate calls for events

William-Unosquare:
Events now get appended when new dashboard event requests are made. A request will be made if a month doesn't have any events, and a request won't be made if events for that month already exist. There's also checks to avoid mandatory holidays blocking requests.

See comments below for more info.


web-app/src/actions/dashboard.js
```diff
@@ -63,30 +64,37 @@ export const setEventView = eventView => {
...
+        })
+        .catch(error => {
+          dispatch(setLoading(false));
+          dispatch(setError(error));
+        });
+      // Fetch team's events as well as your own.
+    } else if (eventView === eventsView.TEAM_EVENTS) {
+      getTeamsEvents(date)
+        .then(({ data }) => {
+          dispatch(setLoading(false));
```
ghost: The fact the code in the block on line 89 onwards plus line 76 onwards is the same....maybe we could pull this out and reuse it within a function?

web-app/src/utilities/dashboardEvents.js
```diff
@@ -49,3 +70,29 @@ export const getEventDuration = event => {
...
+  }, []);
+
+  // Check if any of the dates' month match the month the user is viewing. If
+  // there's a match, we don't want to update.
+  eventDates.forEach(date => {
+    if (date.month() === month) {
+      requireNewRequest = false;
+    }
+  });
+  console.log('reqreq', requireNewRequest);
```
ghost: A log has been left in

ghost (changes requested):
Just a question plus comment

web-app/src/actions/dashboard.js
William-Unosquare: Make sense. Will do. 👍

web-app/src/utilities/dashboardEvents.js
William-Unosquare: Well spotted.

William-Unosquare:
This is working well now I think. This may be worth pulling down and actually giving a test yourselves to make sure I haven't missed any edge cases.

The expected behaviour is:
- Events should be pushed into a state when navigating months.
- When a month is empty, it fires off a request to retrieve events.
- When there are already events for the month being viewed in state, it will not fire a request for events.
- When a user toggles team view or creates an event we force an update. This means all events are removed from state, and we request updated event data for the viewed month.

ghost (approved):

web-app/src/utilities/AxiosInstance.js
```diff
@@ -55,6 +55,19 @@ instance.interceptors.response.use(function(response) {
     };
   }
 
+  if (response.config.url.includes(`${baseURL}/dashboard/getTeamEvents`)) {
```
paulmatchett, line 58: Is this axios middleware required? Does the _formatEventsForCalendar method in dashBoardEvents.js not convert the dates to moments too?

paulmatchett (commented):
Flows great, comments used well and is easy to understand. Only one query and that was regarding the axios middleware for teams call

web-app/src/utilities/AxiosInstance.js
William-Unosquare: Good spot. I think the issue here is actually in the dashboardEvents. I'd rather keep the place we transform dates into moments in the AxiosInstance in case the call was ever used elsewhere in the app. Just keeps it consistent too. 

Pushed a fix.

paulmatchett (approved):

PR #200, opened 2018-08-30
Style enhancements to improve layout and remove sick/work related tra…

mcromie92:
…vel from booking dropdown

### DIFF
--- a/web-app/src/hoc/Layout/nav-menu.js
+++ b/web-app/src/hoc/Layout/nav-menu.js
@@ -13,6 +13,7 @@ import {
 import FontAwesomeIcon from '@fortawesome/react-fontawesome';
 import { faBars, faSignOutAlt } from '@fortawesome/fontawesome-free-solid';
 import Swal from 'sweetalert2';
+import { theme } from './../../styled';
 
 const NavMenu = ({
   history,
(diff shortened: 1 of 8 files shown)


web-app/src/hoc/Layout/nav-menu.js
```diff
@@ -28,6 +28,7 @@ const NavMenu = ({
       type: 'question',
       showCancelButton: true,
       confirmButtonText: 'Yes, log out',
+      confirmButtonColor: '#0eb5d1',
```
ghost: @William-Unosquare  @paul-matchett this is the only thing I wasn't sure on when looking at it with Michaela, obviously the uno hex variable is being used in styled components but this outside of that scope. Not sure how you guys would recommend going forward here. Obviously I think it's fine like this personally.

paulmatchett: Yes, there is no way of accessing the styled components themed props to inject here. So, yes, this is fine

paulmatchett (approved):
All looks good to me.

web-app/src/hoc/Layout/nav-menu.js
William-Unosquare: @mcromie92 @girwin575 You can access the styled components theme by importing theme object directly from the `styled.js` file in the root directory. Then you could do something like `confirmButtonColor: theme.colors.unoBlue`

William-Unosquare (changes requested):
Great stuff! Just take a quick look at the comment above about importing the theme.

mcromie92:
@William-Unosquare, @girwin575 I have updated the CSS for unoBlue importing the theme

William-Unosquare (approved):

William-Unosquare:
Looks good!

PR #201, opened 2018-08-30
Calendar slot selection validation and booking form onchange validation

paulmatchett:
All validation calls have been moved to dashboardEvents utility where they can be shared.
The react-datepicker has also been updated to not allow a user to manually add a date to the input field, as well as restrict weekend dates from being selected.

Calendar Dashboard validation now tests for:

**Using the Calendar or Booking Modal**
Prevent a user from selecting:
- A past date
- Weekend only dates
- Dates that collide with dates already booked, not including event if selected
- Other peoples events

### DIFF
--- a/web-app/src/components/BookingCalendar/container.js
+++ b/web-app/src/components/BookingCalendar/container.js
@@ -3,6 +3,7 @@ import { PropTypes as PT } from 'prop-types';
 import { connect } from 'react-redux';
 import { compose } from 'redux';
 import {
+  validationMessage,
   checkIfPastDatesSelected,
   checkIfDatesFallOnWeekend,
   checkIfSelectedDatesOverlapExisting,
@@ -51,9 +52,9 @@ const BookingCalendarContainer = Wrapped =>
       const pastDatesSelected = checkIfPastDatesSelected(start);
       const datesFallOnWeekend = checkIfDatesFallOnWeekend(start, end);
       if (pastDatesSelected) {
-        return 'Unable to select past dates';
+        return validationMessage.PAST_DATES_SELECTED;
       } else if (datesFallOnWeekend) {
-        return 'Unable to select weekend dates';
+        return validationMessage.WEEKEND_DATES_SELECTED;
       } else {
         const datesOverlapExisting = checkIfSelectedDatesOverlapExisting(
           events,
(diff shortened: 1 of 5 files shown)


web-app/src/utilities/dashboardEvents.js
```diff
@@ -49,3 +50,79 @@ export const getEventDuration = event => {
...
+    if (
+      employee &&
+      employee.employeeId === employeeId &&
+      selectedEventId !== eventId
+    ) {
+      var selectedDateRange = moment.range(
+        moment(start),
+        moment(end).endOf('day'),
+      );
+      var existingEvent = moment.range(moment(event.start), moment(event.end));
```
ghost: I'd try to use let and const for variables.

ghost (changes requested):
Just one comment

web-app/src/components/BookingCalendar/container.js
```diff
@@ -41,52 +46,35 @@ const BookingCalendarContainer = Wrapped =>
...
     };
 
     onSelectSlot = ({ start, end }) => {
-      const validatingDatesResult = this.validateSelectedDates(start, end);
-      if (validatingDatesResult === 'Dates approved') {
+      const calendarValidationResults = this.handleCalendarValidation(
+        start,
+        end,
+      );
+      if (calendarValidationResults === 'Dates approved') {
```
William-Unosquare: I'm still not a huge fan of this error/success message string being used in part of the logic. It's used in quite a lot of places here too.

A lot would break if we wanted to display any other success message to the user.

William-Unosquare (commented):
See comment

web-app/src/components/BookingCalendar/container.js
ghost: Best way to get around this possibly would be a constant file? Strings that are associated with conditions that when changed are changed universally rather than 10 locations. Thoughts guys?

paulmatchett: Some things are totally obvious once pointed out. We've used Const files throughout the application Gary which works well. Let me tidy that up.

web-app/src/utilities/dashboardEvents.js
paulmatchett: Of course. not sure how that slipped in there.

William-Unosquare:
Nice man. Feels much more solid now.

William-Unosquare (approved):

PR #204, opened 2018-09-03
Feature/styling updates

mcromie92:


web-app/src/pages/TeamDashboard/styled.js
```diff
@@ -3,11 +3,13 @@ import styled from 'styled-components';
...
 export const Stat = styled.div`
-  max-width: 200px;
+  max-width: 100%;
+  flex: 1;
+  margin-right: 15px;
   h1 {
     margin-bottom: 5px;
     color: ${props => props.theme.colours.unoBlue};
-  }
+    }
```
ghost: Small issue, tabbing has been indented a little too much.

web-app/src/pages/TeamDashboard/styled.js
```diff
@@ -17,7 +19,7 @@ export const Columns = styled.div`
   @media (min-width: ${props => props.theme.mediaQueries.lg}) {
     justify-content: space-around;
     flex-direction: row;
-  }
+    }
```
ghost: Same again here for tabbing

web-app/src/pages/Profile/index.jsx
```diff
@@ -30,15 +32,15 @@ const Profile = props => {
           <div className="columns">
             <div>
               <h1>{totalHolidays - daysBooked}</h1>
-              <h4>Days Remaining</h4>
+              <h4><FontAwesomeIcon icon={faCalendar} /> Days Remaining</h4>
```
ghost: So in my eyes here it should be `<FontAwesomeIcon icon={faCalendar} /><h4>Days Remaining</h4>`

Easier to read and the header tag should only contain the header text. Same will apply for the others below.

ghost (changes requested):
A few changes here.

web-app/src/pages/Profile/index.jsx
ghost: Just looking through the other files, it seems this has been done across the app. It would be cleaner to have it `<FontAwesomeIcon icon={faCalendar} /><h4>Days Remaining</h4>` but because it's been done already just leave this comment out.

paulmatchett: Its easier to have the font inside the h4 tag so that the icons inherits their styles. Possibly be tidier if the text was wrapped in a span tag? So it would be like this:

```
<h4>
  <FontAwesomeIcon icon={faCalendar} />
  <span>Days Remaining</span>
</h4>
```
Although if it's not done like that anywhere else, I agree with Gary, keep as is.

ghost: Issue being it's not valid HTML but it still works of course. I wouldn't put the span in either. It ideally shouldn't have nested tags but I'd keep it how it is as it's working and done across the app.

mcromie92:
@girwin575 removed the tab indents- thanks

web-app/src/pages/Profile/index.jsx
```diff
@@ -30,15 +32,15 @@ const Profile = props => {
...
             </div>
             <div>
               <h1>{daysBooked}</h1>
-              <h4>Days Booked</h4>
+              <h4><FontAwesomeIcon icon={faCheck} /> Days Booked</h4>
```
William-Unosquare: To keep this even more consistent with the team page append the word "Days" on to the end of the H1 number and remove it from the H4.

William-Unosquare (changes requested):
One small change but it looks great otherwise!

mcromie92:
@William-Unosquare changes made to make consistent with team page

William-Unosquare:
Sweet! I think this is good to merge.

William-Unosquare (approved):

PR #208, opened 2018-09-03
Feat/improve calendar design

kirsty-unosquare:
Redesign of calendar and added new upcoming list. It's not complete but nearly, just didn't have enough time to finish it. It all works but the list item could be moved to the common folder since it's already being used.

This branch might conflict with Lee's not sure if it's already been merged.


NativeApp/src/components/Home/container.js
```diff
@@ -67,6 +68,8 @@ export default Container => class extends Component {
       const { calendarDate } = this.state;
       getMonthEvents(calendarDate)
         .then(data => this.setState({ events: this.formatDate(data) }));
+      getUserEvents()
+        .then(res => this.setState({ upcomingEvents: res }));
```
d1wilko, line 72: could make this... `.then(upcomingEvents => this.setState({ upcomingEvents }));` - but no big

d1wilko (approved):
nice code - one suggestion but don't mind if you take it in onboard or not :)

PR #209, opened 2018-09-04
Feature/add contracts to profile

mcromie92:
Some files have overlapped from my other branch, please merge previous branch first

### DIFF
--- a/web-app/src/pages/Profile/container.js
+++ b/web-app/src/pages/Profile/container.js
@@ -10,12 +10,14 @@ import {
 } from '../../utilities/dates';
 import holidayStatus from '../../utilities/holidayStatus';
 import { isEmpty } from 'lodash';
+import { getContractsByEmployeeId } from '../../services/contractService';
+import swal from 'sweetalert2';
 
 const ProfileContainer = Wrapped =>
   class extends React.Component {
     static propTypes = {
       userDetails: PT.object.isRequired,
-    };
+   };
 
     constructor(props) {
       super(props);
@@ -24,12 +26,14 @@ const ProfileContainer = Wrapped =>
         daysBooked: null,
         daysPending: null,
         selectedHoliday: {},
+        contracts: [],
       };
     }
 
     componentDidMount() {
       // Required in case the user navigates away from the page, then back.
       this.setState({ holidays: null });
+      this.setState( { contracts: [] });
     }
 
     componentDidUpdate(prevProps, prevState) {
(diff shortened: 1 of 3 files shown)


web-app/src/pages/Profile/container.js
```diff
@@ -56,7 +60,20 @@ const ProfileContainer = Wrapped =>
             });
           });
         });
-      }
+      } 
+      if (
+        this.props.userDetails.employeeId !== null &&
+        this.state.contracts !== null
+      )
+        {
```
ghost: Tabbing seems to be a bit off.

web-app/src/pages/Profile/container.js
```diff
@@ -24,12 +26,14 @@ const ProfileContainer = Wrapped =>
...
         daysPending: null,
         selectedHoliday: {},
+        contracts: [],
       };
     }
 
     componentDidMount() {
       // Required in case the user navigates away from the page, then back.
       this.setState({ holidays: null });
+      this.setState( { contracts: [] });
```
ghost: I'd just make sure the spacing etc is the same as the others.

web-app/src/pages/Profile/container.js
ghost: So what happens if the conditions fails, it just doesn't run `getContractsByEmployeeId` is that expected behaviour?

ghost (changes requested):
Just a few things, tabbing needs just to be looked at closer.

web-app/src/pages/Profile/container.js
```diff
@@ -56,7 +60,20 @@ const ProfileContainer = Wrapped =>
...
           });
         });
-      }
+      } 
+      if (
+        this.props.userDetails.employeeId !== null &&
+        this.state.contracts !== null
+      )
+        {   
+        getContractsByEmployeeId(this.props.userDetails.employeeId).then(response => {
```
William-Unosquare: We already chatted about this, but I'm just making sure this isn't merged in yet by accident. Basically, we can't update state in `ComponentDidUpdate` or you'll be stuck in a loop.

William-Unosquare (changes requested):

William-Unosquare:
As for tabbing, if you're using VSCode, get the extension [prettier](https://prettier.io/). It'll keep your code style in check automatically every time you save the file. It's a great time saver.

You'll also need to set these preferences up (File -> Preferences -> Settings)
```
{
  "editor.tabSize": 2,
  "prettier.trailingComma": "all",
  "prettier.singleQuote": true,
  "prettier.eslintIntegration": true,
}
```

mcromie92:
@girwin575 fixed the tabbing and spacing, installed prettier now too :)
@William-Unosquare I have updated the function so it is not called inside the componentDidUpdate to remove the loop

William-Unosquare (approved):
Looks great, approved!

PR #212, opened 2018-09-06
Feature/ts 31 - As an admin I can reject an event request and as a user I can view the rejected event and the reason why

eoinmcafee:

PR #213, opened 2018-09-10
Fix/ts 46 dash colours

William-Unosquare:
Relies on back-end from TS-31 (#212)

- Fixes issue with event background colours not matching up with statuses
- Fixes additional issues where events couldn't be created or updated because back-end now expects `eventId` in the request instead of `holidayId`.
- Adds "Annual Leave" type to filtering, in case you don't want to see working from home events.
- Updated icons


d1wilko (approved):
looks grand to me

PR #216, opened 2018-09-11
[Task TS-30] Back end validation for number of holidays for employee …

mccleaveryan:
…& no duplication of existing events with the same dates.

Branch not ready for merging into develop. 
Review needed for back end validation approach.

Number of holidays can not be below 0.
An event cannot be created if an event is created by an employee with the same dates.


BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/EventRepository.java
```diff
@@ -15,9 +18,26 @@
...
+    List<Event> findByStartDateBetweenAndEventType(LocalDate rangeStart, LocalDate rangeEnd, EventType eventType);
 
     List<Event> findByEventStatusAndEventType(EventStatus eventStatus, EventType eventType);
 
     List<Event> findByEventType(EventType eventType);
+
+    @Query(value = "SELECT totalHolidays " +
+            "FROM Employee " +
+            "WHERE employeeId = :employeeId")
+    Long getTotalEventsRemainingByEmployee(@Param("employeeId") int employeeId);
```
BrownMark: Does this belong on the event repository? We're querying from employee table

BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/EventRepository.java
```diff
@@ -15,9 +18,26 @@
...
+
+    @Query(value = "SELECT COUNT(eventType) " +
+            "FROM Event " +
+            "WHERE " +
+            "employee = :employeeId " +
+            "AND " +
+            "startDate = :startDate " +
+            "AND " +
```
BrownMark: I'm not sure if this is the correct approach, surely it needs to intersect? for example if we have 2 ranges 1st - > 3rd and 2nd -> 4th it would not be caught here.

BackEnd/src/main/java/com/unosquare/admin_core/back_end/controller/HolidayController.java
```diff
@@ -71,6 +71,22 @@ public ResponseEntity createHoliday(@RequestBody CreateHolidayViewModel createHo
                 responses.add("Starting date cannot be after end date");
                 continue;
             }
+
+            Long numberOfRemainingHolidays = eventService.getNumberOfHolidaysRemainingForEmployee(
+                    createHolidayViewModel.getEmployeeId());
+
+            if (numberOfRemainingHolidays == 0){
```
BrownMark: What about if the user has 1 remaining day, but requests 3?

BrownMark (changes requested):

BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/EventRepository.java
mccleaveryan: Changed to event repository there. Didn't notice I had it in the wrong repo

BrownMark: One thing to keep in mind, this number does not decrement. It is a set value that represents the available holidays for the year.

the holidays remaining = totalholidays - (sum of holiday events).

mccleaveryan: Ok but is there a value that decrements each time an event is created?

BackEnd/src/main/java/com/unosquare/admin_core/back_end/controller/HolidayController.java
```diff
@@ -57,8 +67,69 @@ public HolidayViewModel findHolidayById(@PathVariable("eventId") int eventId) {
 
     @PostMapping(value = "/", produces = MediaType.APPLICATION_JSON_VALUE)
     @ResponseStatus(HttpStatus.CREATED)
-    public ResponseEntity<List<String>> createHoliday(@RequestBody CreateEventViewModel createEventViewModel) {
-        List<String> responses = createEventByType(createEventViewModel, EventTypes.ANNUAL_LEAVE);
+    public ResponseEntity createHoliday(@RequestBody CreateEventViewModel createHolidayViewModel) {
+
```
eoinmcafee, line 71: The majority of this code lives in the base controller now. Anything new you've added needs to go in there instead. If not delete it out. Have you pulled the latest develop?

BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/DashboardRepository.java
```diff
@@ -47,7 +47,8 @@
                                                @Param("endDate") LocalDate endDate,
                                                @Param("today") LocalDate today);
 
-    @Query(value = "SELECT new com.unosquare.admin_core.back_end.dto.EmployeeSnapshotDto(t.teamId ,CONCAT(emp.forename, ' ', emp.surname), et.description, t.teamName) " +
+    @Query(value = "SELECT new com.unosquare.admin_core.back_end.dto.EmployeeSnapshotDto(t.teamId ," +
```
eoinmcafee: Again this query has changed since your last pull from develop. It looks like you didn't make any code changes to this class other than formatting. It's best practice in this instant to reset it back to the original. No point in introducing unnecessary changes to a file. Adds an unnecessary history record to the file

BackEnd/src/main/java/com/unosquare/admin_core/back_end/repository/EventRepository.java
```diff
@@ -13,11 +15,44 @@
...
+//                    "FROM Event " +
+//                    "WHERE " +
+//                    "employee_id = :employeeId " +
+//                    " AND " +
+//                    "(start_date >= '2018-01-01') AND " +
+//                    "(end_date <= '2018-12-31') " +
+//                    ")totalEvents", nativeQuery = true)
+//    Long getCountOfTotalEventsInYearMadeByEmployee(@Param("employeeId") Employee employee);
+
+    @Query(value = "SELECT e.employee.employeeId " +
```
eoinmcafee, line 54: Do you need just need an employeeId at this point ? What is it being used for? If you are then using that Id to pull back an employee it would be more efficient to just return the full employee entity? Again i'm not sure if this is the use case?

mccleaveryan: I'm not sure which query you're referencing here. But if it's for the 'Admin/Team leads can not approve their own holidays', having back just the 'employeeID' that's on the specific event/holiday would surely be enough. Rather than pull back a full employee entity. If we are using the 'employeeCredentialsViewModel.getUserId()' on the holiday controller then comparing the two ID's is simply all that's needed.

PR #218, opened 2018-09-11
[TS-42] Fix loading overlay causing flicker on fast connections

William-Unosquare:
If you had a fast connection and data loaded really fast, the loading overlay would create what looked like a page flicker.

Now the indicator only displays if the user has been waiting for data for more than 1 second. This means slow connections will still see the indicator but fast ones will have data come in smoothly creating a nicer experience.

### DIFF
--- a/web-app/src/actions/dashboard.js
+++ b/web-app/src/actions/dashboard.js
@@ -1,7 +1,7 @@
 import * as actionTypes from '../actionTypes';
 import { getUsersEvents, getTeamsEvents } from '../services/dashboardService';
 import eventsView from '../utilities/eventsView';
-import { setLoading } from './loading';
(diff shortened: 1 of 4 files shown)


ghost (approved):
Nice move.

William-Unosquare:
@BrownMark Suggested that I also add in a minimum display time for the loading indicator so it never flashes in edge-cases like if the latency was 1011ms.

So now the loading indicator doesn't display until 500ms has past meaning fast connections still won't see any indicator. However, if 500ms does pass then the indicator will display for a minimum of one second.

Could you take another quick look at this @girwin575 ?

ghost:
Looks good @William-Unosquare nicer approach!

PR #219, opened 2018-09-11
[TS-53] Stop overlapping and responsive styling for event type toggle

William-Unosquare:
- Fixes overlapping of event type toggle & calendar navigation buttons.
- Improves responsiveness of event type toggle.

chrome_2018-09-11_16-36-11

chrome_2018-09-11_16-36-19

chrome_2018-09-11_16-36-28

### DIFF
--- a/web-app/src/pages/Dashboard/styled.js
+++ b/web-app/src/pages/Dashboard/styled.js
@@ -2,11 +2,31 @@ import styled from 'styled-components';
 
 export const ButtonToggle = styled.div`
   position: absolute;
-  left: calc(60% - 111px);
+  right: 350px;
   top: 30px;
-  display: flex;
   height: 40px;
-   `;
+  display: none;
+  .btnText {
+    display: none;
+  }
+
+  span {
+    margin: 0 !important;
+  }
+
+  @media (min-width: ${props => props.theme.mediaQueries.lg}) {
+    .btnText {
+      display: inline;
+    }
+    span {
+      margin: inherit;
+    }
+  }
+
+  @media (min-width: ${props => props.theme.mediaQueries.md}) {
+    display: flex;
+  }
+`;
 
 export const InnerLayout = styled.div`
   width: 100%;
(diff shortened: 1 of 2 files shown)


mcromie92 (approved):
looks good

ghost:
@William-Unosquare just looking at the screenshot. Any chance a bit of spacing could be added between the icon and the text 'Team Events'

ghost (changes requested):
Just about the spacing for the icon and Team events text.

web-app/src/pages/Dashboard/styled.js
```diff
@@ -2,11 +2,31 @@ import styled from 'styled-components';
...
   top: 30px;
-  display: flex;
   height: 40px;
-   `;
+  display: none;
+  .btnText {
+    display: none;
+  }
+
+  span {
```
ghost: any chance we could not use !important?

ghost: Small thing but try to give everything a class rather than put CSS on a element.

bgribz: I don't think this is an issue when used with styled-components, maybe even preferable to keep code clean.

ghost: It's not a stylistic reason, it's just for performance. A small thing one time doesn't make a huge difference but 100 times makes all the difference across an app. Appreciate it's CSS within JS but it should still stand. I've been too quite a few performance talks now and it's always brought up with CSS. No tags and try to stay away from ID's

ghost: Go with whatever is consistent to what already has been done but it's just good to note for future.

ghost: @bgribz was beside me and mentioned he thinks a class is generated. I checked https://www.styled-components.com/docs/advanced and inspected a box and it does generate a class. So ignore what I said there in terms of styled components but the principle stands

ghost (approved):

William-Unosquare:
Cleaned up the CSS. I used class names because it came with the bonus that it solved the !important along with a cascade swap. Also changed btnText to btn-text because it's how the other classes in the file are named.

Also fixed the margin on the icon. Ooops.

PR #226, opened 2018-09-14
[TS-67] Fix ability to cancel holidays by calling correct endpoint.

William-Unosquare:
Cancelling a holiday now calls the new `cancelHoliday` endpoint. I've added a temporary solution to confirming a holiday cancellation that should be changed later. I didn't want to modify the file too much because Paul is working in the same area of code.

### DIFF
--- a/web-app/src/components/BookingModal/ModalStatusBanner/index.jsx
+++ b/web-app/src/components/BookingModal/ModalStatusBanner/index.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
 import { PropTypes as PT } from 'prop-types';
 import { Banner } from './styled';
 import eventCategory from '../../../utilities/eventCategory';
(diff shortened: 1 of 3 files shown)


mcromie92 (approved):
looks good :)

BrownMark (approved):

web-app/src/components/BookingModal/ModalStatusBanner/index.jsx
```diff
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { Component } from 'react';
```
PaulTummonUnosquare, line 1: Just a small thing that if your extracting React already could you just use React.Component to make the class

PaulTummonUnosquare (commented):
Yeah it all looks good!  : )

web-app/src/components/BookingModal/ModalStatusBanner/index.jsx
William-Unosquare: This way of importing Component is generally the best practice.

PR #228, opened 2018-09-17
[TS-70] Legend now appears on the right of the calendar, instead of below

William-Unosquare:
I made this change for a few reasons:
- It removes the need to scroll up and down the page to interact with the filters.
- This layout makes better use of space. Previously keys would have appeared really wide for no reason.
- Allows for the calendar to have more height and therefore fit more events in view.

chrome_2018-09-17_14-07-27

### DIFF
--- a/web-app/src/pages/Dashboard/styled.js
+++ b/web-app/src/pages/Dashboard/styled.js
@@ -2,7 +2,7 @@ import styled from 'styled-components';
 
 export const ButtonToggle = styled.div`
   position: absolute;
-  right: 350px;
+  right: 575px;
   top: 30px;
   height: 40px;
   display: none;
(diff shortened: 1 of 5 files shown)


web-app/src/pages/Dashboard/styled.js
```diff
@@ -50,6 +50,23 @@ export const InnerLayout = styled.div`
...
+    border-radius: 5px 5px 0 0;
+    overflow: hidden;
+  }
+
+  .rbc-month-view {
+    border: none;
+  }
+
+  .rbc-month-row {
+    border-left: 1px solid #ddd;
```
ghost: variable for #ddd?

ghost (changes requested):
Small comment

d1wilko (approved):
looks good to me :)

ghost (approved):

PR #231, opened 2018-09-18
[TS-74] Add loading state to tables in profile page 

William-Unosquare:
When data is being fetched the tables will now display as loading.

### DIFF
--- a/web-app/src/components/DataTable/index.jsx
+++ b/web-app/src/components/DataTable/index.jsx
@@ -69,7 +69,6 @@ class DataTable extends Component {
       acc.push(column.Header);
       return acc;
     }, []);
-    console.log(loading);
 
     if (data.length < 0 && !loading) {
       return <p>{emptyMessage}</p>;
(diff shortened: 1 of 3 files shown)


web-app/src/components/DataTable/index.jsx
```diff
@@ -59,15 +62,17 @@ class DataTable extends Component {
...
     const formattedColumns = this.buildColumns(columns);
 
     const labels = formattedColumns.reduce((acc, column) => {
       acc.push(column.Header);
       return acc;
     }, []);
 
-    if (data.length > 0) {
+    if (data.length < 0 && !loading) {
+      return <p>{emptyMessage}</p>;
```
ghost, line 74: emptyMessage?

William-Unosquare: It’s the message that’s displayed when there’s no data. May get rid of it honestly and just show an empty table instead.

ghost: Ah I see, yeah whatever you think is better. For a second I was wondering whether it was just empty that variable haha.

ghost (approved):

web-app/src/components/DataTable/index.jsx
William-Unosquare: I'll keep it like this for now just so I can work on more MVP critical issues.

PR #235, opened 2018-09-19
section list for user profile

lee-cheung:


NativeApp/src/components/User/View/index.js
```diff
@@ -44,6 +42,17 @@ const UserView = (props) => {
...
                 />
               )}
+              renderSectionHeader={({ section }) => (
+                section.data.length >= 1
+                  ? (<H3 type="bold" style={styles.sectionListHeader}>{section.title}</H3>)
+                  : (
+                    <View>
+                      <H3 type="bold" style={styles.sectionListHeader}>{section.title}</H3>
+                      <H3 style={styles.sectionListNoItems}>Nothing to show</H3>
+                    </View>)
```
William-Unosquare: Is there anywhere else you can display the "nothing to show"? Feels weird to be sticking a div and 2 h3's inside 1 header.

NativeApp/src/components/User/container.js
```diff
@@ -46,6 +47,25 @@ export default Container => class extends Component {
     this.sub.remove();
   }
 
+  sortingEvents = (events) => {
+    let eventArray = [];
+    let approvalArray = [];
+    let awaitApprovalArray = [];
+
+    approvalArray.push(events.filter(event => event.eventStatus.description === 'Approved'));
+    awaitApprovalArray.push(events.filter(event => event.eventStatus.description === 'Awaiting approval'));
```
William-Unosquare: I'd call these something more like `approvedEvents` and `pendingEvents`. If "awaitApproval" is the terminology you've used elsewhere in the app then ignore me. :)

William-Unosquare (changes requested):
Question & a wee possible change.

NativeApp/src/components/User/View/index.js
bgribz: I kinda think that multiline ternaries like these are a bit untidy. Another approach would be to move the logic to a function that will return either component, or ideally do this kind of logic in a container that will pass the correct component through on props.

bgribz (changes requested):
Just a small suggestion

NativeApp/src/components/User/View/index.js
lee-cheung: i have made changes as per you suggestions. please could you check it out? thanks :)

bgribz (approved):

William-Unosquare (approved):
Looks good

PR #237, opened 2018-09-20
holiday rejection messaging functionality

PaulTummonUnosquare:


web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -75,6 +74,13 @@ const BookingModalForm = props => {
     }
   };
 
+  const renderRejectionReasonMessage = booking => {
+    if (booking.messages) {
+      return false;
+    }
+    return true;
+  };
```
William-Unosquare: This seems like a very long-winded way to check the value of a bool.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -118,6 +124,16 @@ const BookingModalForm = props => {
           }}
           label="End Date:"
         />
+        <Input
+          type="input"
+          htmlAttrs={{
+            type: 'input',
+            name: 'employeeRejectionMessage',
+            disabled: renderRejectionReasonMessage(booking),
```
William-Unosquare: `disabled: !booking.messages` would get rid of that long-winded bool check func

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -153,10 +169,11 @@ BookingModalForm.propTypes = {
...
   updateEvent: PT.func.isRequired,
+  booking: PT.object,
 };
 
 BookingModalForm.defaultProps = {
   formIsValid: true,
 };
 
-export default container(BookingModalForm);
+export default container(BookingModalForm);
```
William-Unosquare: Not a big issue but its seems like your Prettier still isn't adding newlines at the end of the file.

web-app/src/components/BookingModal/index.js
```diff
@@ -4,8 +4,17 @@ import ModalStatusBanner from './ModalStatusBanner';
 import { PropTypes as PT } from 'prop-types';
 import BookingModalForm from './BookingModalForm';
 import { Modal } from '../common';
+
 import { StyleContainer, FormContainer } from './styled';
 
+const rejectionReason = booking => {
+
```
William-Unosquare: spacing

web-app/src/components/HolidayModal/container.js
```diff
@@ -79,24 +102,60 @@ const HolidayModalContainer = Wrapped =>
...
         });
     };
 
+    collapseRejectionReasonState = () => {
+      this.setState({ holidayModalExpansion: false });
+    };
+
+    expandRejectHolidayExplanationText = () => {
+      this.setState({ holidayModalExpansion: true });
+    };
```
William-Unosquare: These could be combined together and pass true or false as a parameter.

web-app/src/components/HolidayModal/index.js
```diff
@@ -26,7 +31,14 @@ const HolidayModal = ({
...
   };
 
+  const shouldDisableButtonControl = inputText => {
+    return !inputText.length > 0;
+  };
+
   const duration = getEventDayAmount(holiday);
+  const disableRejectionReasonButton = shouldDisableButtonControl(
+    capturedRejectionReasonText
+  );
```
William-Unosquare: Maybe just me, but I see this as additional complexity that's unneeded.

`disabled={inputText.length === 0}` is still readable and is far less code.

William-Unosquare (changes requested):
I'm sure you don't want to hear this, but another few changes to make here. Getting there though!

web-app/src/components/BookingModal/styled.js
```diff
@@ -64,3 +63,9 @@ export const FormContainer = styled.div`
     }
   }
 `;
+
+export const ButtonFloat = styled.div`
+  width: 25% !important;
+  margin-bottom: 10px !important;
+  float: right;
+`;
```
ghost: New line here

web-app/src/components/common_styled/InputText.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components';
+
+export default styled.input`
+  font-family: inherit;
+  font-size: 100%;
+  line-height: 1.15;
+  margin: 10px 0px 10px 0px;
+  height: 34px;
+  width: 100%;
+`;
```
ghost: New line

web-app/src/services/holidayService.js
```diff
@@ -24,10 +24,13 @@ export const approveHoliday = eventId => {
   return axios.put('/holidays/approveHoliday', { eventId });
 };
 
-export const rejectHoliday = eventId => {
+export const rejectHoliday = (eventId, message) => {
+  if (message) {
+    return axios.put('/holidays/rejectHoliday', { eventId, message });
+  }
   return axios.put('/holidays/rejectHoliday', { eventId });
```
girwin575: Not sure I like a check on the message to be honest, my personal preference should be if no message is sent there it's default set to empty. The backend should do the crux of the work after that, we just send them the details. Happy to discuss if anyone doesn't agree. 

`export const rejectHoliday = (eventId, message='') => {
  return axios.put('/holidays/rejectHoliday', { eventId, message });
};`

web-app/src/utilities/dashboardEvents.js
```diff
@@ -173,4 +174,4 @@ export const checkIfSelectedDatesOverlapExisting = (
     }
   });
   return overlappingEvents.length > 0;
-};
+};
```
ghost, line 177: new line

web-app/src/services/holidayService.js
girwin575: Or we could handle this upstream a bit more, in the actual component when passing the data through.

girwin575: @eoinmcafee I'd like your thoughts on this also above.

web-app/src/services/holidayService.js
```diff
@@ -24,10 +24,13 @@ export const approveHoliday = eventId => {
   return axios.put('/holidays/approveHoliday', { eventId });
 };
 
-export const rejectHoliday = eventId => {
+export const rejectHoliday = (eventId, message) => {
+  if (message) {
```
ghost: Thinking a reject object would be a lot cleaner here.

`{
  eventId: 1,
  message: 'rejection message here'
}`

Maintainability plus a cleaner pure service here. No logic should exist in the services, it takes the parameters and hits the endpoint. @eoinmcafee would it be ok your side, to take an object instead of two strings and pull the values out of the object.

ghost (dismissed):
A few ideas and thoughts plus some small styling issues.

web-app/src/services/holidayService.js
PaulTummonUnosquare: Inspecting this method being called, there is reference to rejectHoliday elsewhere but it is a different function completely so the only place that calls rejectHoliday to hit that endpoint is the modal which always requires a rejection message so I have removed the logic and sending the message at all times as it should always be being passed in.

ghost: Fair enough. As long as the service is clean.

William-Unosquare (approved):
Fix the new lines and this is good to go I think

ghost (approved):

PR #238, opened 2018-09-20
[TS-51][TS-76][TS-75] Team page rebuild for supporting multiple teams & daily snapshot

William-Unosquare:
This PR is a rebuild of the old team page with some old stuff carried over. It's built to work off the new team snapshot endpoint. This means:
- The page now supports users who are in multiple teams with a new sidebar.
- The statistics at the top of the page are now no longer hard-coded.
- The table now displays "Today's status", indicating where that user is on the current day.

Notes:
- Pending holidays has been removed so we can support it with an API call just for that after MVP (saying only admins will be approving holidays to start with).
- The statistics that appear about the user are hacked together right now. We also need an API call for that, but it's not critical so again, this can wait until after MVP.
- The client is mocked right now. Awaiting the back-end lads to add the client to the snapshot api call.

image


web-app/src/pages/TeamDashboard/index.js
```diff
@@ -1,114 +1,102 @@
...
-              columns={['employee', 'startDate', 'endDate']}
-              onRowClick={holiday => selectHoliday(holiday)}
-            />
-          </div>
-        )}
-      </Columns>
-    </Container>
+      <Layout>
+        <TeamSidebar teams={teams} selectTeam={selectTeam} />
+        {team ? renderTeamDetails() : null}
```
bgribz: `team && renderTeamDetails()`

bgribz (changes requested):

web-app/src/components/UserModal/container.js
```diff
@@ -24,27 +24,31 @@ const UserModalContainer = Wrapped =>
...
-      }
+
+      getHolidays(user.employeeId)
+        .then(response => {
+          if (response.data.length > 0) {
+            this.setState({
+              totalHolidays: response.data[0].employee.totalHolidays,
+            });
+          }
+          this.setState({ userHolidays: response.data });
```
bgribz: Rather than having (potentially) two sequential setStates, would it be worth having this as 
```
this.setState({
  userHolidays: response.data,
  totalHolidays: response.data.length > 0 ? response.data[0].employee.totalHolidays : 0,
})
```

bgribz (changes requested):

bgribz (approved):

PR #244, opened 2018-09-21
added client name to snapshot calls and fixed bug with event messages

eoinmcafee:


BackEnd/src/main/java/com/unosquare/admin_core/back_end/controller/DashboardController.java
```diff
@@ -98,6 +98,7 @@ private void mapEmployeeSnapshotViewModel(EmployeeSnapshotDto m, EmployeeSnapsho
         employee.setName(m.getName());
         employee.setEmployeeId(m.getEmployeeId());
         employee.setEmail(m.getEmail());
+        employee.setClientName(m.getClientName());
```
BrownMark, line 101: Can this mapper be completed with modelmapper? Just so we can keep our mapping logic separated

BrownMark (approved):

PR #247, opened 2018-09-24
fix for contract date picker

PaulTummonUnosquare:


web-app/src/components/CreateContract/CreateContractForms/dates/index.jsx
```diff
@@ -33,19 +32,18 @@ export const DateForm = props => {
...
+          label="End Date:"
+          type="date"
+          htmlAttrs={{
+            type: 'input',
+            name: 'endDate',
+            placeholder: 'Enter a end date',
+            disabled: formData.isOpenEnded,
+          }}
+          value={formData.endDate}
+          rules={{}}
```
ghost, line 45: Just curious as I don't know the answer, does rules have to be there since it's an empty object? It's expected I assume, just checking :)

ghost (approved):

web-app/src/components/CreateContract/CreateContractForms/dates/index.jsx
PaulTummonUnosquare: Not sure as this that prop being passed in was already there. I assume it requires an empty object to full-fill it but will have a look into whether it can be removed

PR #255, opened 2018-09-27
added filter functionality on incoming new event format object for sp…

PaulTummonUnosquare:
…litting holiday bookings over across weekends

### DIFF
--- a/web-app/src/utilities/AxiosInstance.js
+++ b/web-app/src/utilities/AxiosInstance.js
@@ -7,12 +7,12 @@ const instance = axios.create({
   baseURL,
 });
 
-instance.interceptors.request.use(function(config) {
+instance.interceptors.request.use(function (config) {
   config.headers.Authorization = `Bearer ${localStorage.getItem('id_token')}`;
   return config;
 });
 
-instance.interceptors.response.use(function(response) {
+instance.interceptors.response.use(function (response) {
   // Adds start, end & requested to response which are in moment format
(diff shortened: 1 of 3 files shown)


web-app/src/utilities/AxiosInstance.js
```diff
@@ -36,22 +36,27 @@ instance.interceptors.response.use(function(response) {
...
     const employee = store.getState().USER;
-    for (let event of events) {
-      // Raw dates to moment objects
-      event.start = new moment(event.startDate, 'YYYY-MM-DD');
-      event.end = new moment(event.endDate, 'YYYY-MM-DD');
-      // Append logged in employee
-      event.employee = { ...employee };
-    }
+    const arrayOfEvents = [];
+    events.map(event => {
```
ghost, line 45: Happy to go with this for now knowing that the backend will change the data structure at a certain point.

ghost (approved):

web-app/src/utilities/AxiosInstance.js
```diff
@@ -36,22 +36,27 @@ instance.interceptors.response.use(function(response) {
...
+      event.events.map(nestedEvent => {
+        const eventId = nestedEvent.eventId ? nestedEvent.eventId : eventId;
```
William-Unosquare, line 47: When you're assigning the false `eventId` here in the ternary, does it exist? I can't find anywhere where it's assigned previous to this line.

PR #261, opened 2018-10-01
styling header navigation

lee-cheung:


NativeApp/src/screens/Booking.js
```diff
@@ -2,12 +2,20 @@ import React, { Component } from 'react';
...
     title: 'Booking',
     headerBackTitle: null,
-    headerBackImage: () => <Icon type="font-awesome" name="arrow-left" size={20} color="grey" />,
+    headerTitleStyle: {
+      color: WHITE,
+      marginHorizontal: 0,
+      fontSize: H2_SIZE,
+      fontFamily: 'oswaldRegular',
+    },
+    headerLeft: <Icon containerStyle={{ paddingLeft: 15 }} type="font-awesome" name="arrow-left" size={20} color="white" />,
```
bgribz: This would be more readable with the props split onto multiple lines
```
headerLeft: 
  <Icon 
    containerStyle={{ paddingLeft: 15 }} 
    type="font-awesome" 
    name="arrow-left" 
    size={20} 
    color="white" 
  />,
```
or whatever compiles lol

NativeApp/src/utilities/styles.js
```diff
@@ -0,0 +1,32 @@
...
+  headerTitleStyle: {
+    color: WHITE,
+    fontSize: H2_SIZE,
+    fontFamily: 'oswaldRegular',
+  },
+  headerStyle: {
+    paddingHorizontal: 5,
+    backgroundColor: UNOBLUE,
+    borderBottomWidth: 1,
+    borderBottomColor: 'rgba(192,192,192,0.3)',
```
bgribz, line 16: Could this become a common colour or is this the only place it's used?

bgribz (changes requested):
2 small queries but otherwise great 👍

NativeApp/src/screens/Booking.js
lee-cheung: I have put it back to headerBackImage as the back button does not work without it

NativeApp/src/utilities/styles.js
lee-cheung: This is the only place it is used.

William-Unosquare (approved):

bgribz (approved):

PR #267, opened 2018-10-03
Adding a message to stop users from booking more days off than what t…

mcromie92:
…hey have remaining
> calculated the approved days
>caculated the total days
> from this calculated the remaining holidays user has left 
> show message if the user tries to book more days than what they have allocated 
capture1

### DIFF
--- a/web-app/src/components/BookingModal/BookingModalForm/container.js
+++ b/web-app/src/components/BookingModal/BookingModalForm/container.js
@@ -81,6 +81,13 @@ const Container = Wrapped =>
         HolidayStatus.APPROVED
       );
     };
+
+    getPendingDays = () => {
+      return getTotalDaysInEventArrayWithStatus(
+        this.props.allEvents,
+        HolidayStatus.PENDING
+      );
+    };
     
     getTotalDays = (this.props.userDetails.totalHolidays);
 
@@ -182,7 +189,8 @@ const Container = Wrapped =>
           toggleRejectionMessageView={toggleRejectionMessageView}
           capturedRejectionReponseText={capturedRejectionReponseText}
           assignRejectionResponseText={this.assignRejectionResponseText}
-          getApprovedDays={this.getApprovedDays}
+          getApprovedDays={this.getApprovedDays()}
+          getPendingDays={this.getPendingDays()}
           toggleRejectionResponseView={toggleRejectionResponseView}
           isEventBeingUpdated={isEventBeingUpdated}
           bookingDuration={bookingDuration}
--- a/web-app/src/components/BookingModal/BookingModalForm/container.js
+++ b/web-app/src/components/BookingModal/BookingModalForm/container.js
@@ -15,6 +15,8 @@ import {
 import { Toast } from '../../../utilities/Notifications';
 import moment from 'moment';
 import { getUser, getAllEvents, eventBeingUpdated } from '../../../reducers';
+import { getTotalDaysInEventArrayWithStatus } from '../../../utilities/dates';
+import HolidayStatus from '../../../utilities/holidayStatus';
 
 const Container = Wrapped =>
   class extends React.Component {
@@ -49,6 +51,7 @@ const Container = Wrapped =>
           start: moment(),
           employeeRejectionMessage: '',
           updateMessage: '',
+          totalHolidays: 0,
         },
         formIsValid: true,
         capturedRejectionReponseText: '',
@@ -71,6 +74,15 @@ const Container = Wrapped =>
         },
       });
     };
+
+    getApprovedDays = () => {
+      return getTotalDaysInEventArrayWithStatus(
+        this.props.allEvents,
+        HolidayStatus.APPROVED
+      );
+    };
+    
+    getTotalDays = (this.props.userDetails.totalHolidays);
 
     assignRejectionResponseText = e => {
       this.setState({ capturedRejectionReponseText: e.target.value });
@@ -170,6 +182,7 @@ const Container = Wrapped =>
           toggleRejectionMessageView={toggleRejectionMessageView}
           capturedRejectionReponseText={capturedRejectionReponseText}
           assignRejectionResponseText={this.assignRejectionResponseText}
+          getApprovedDays={this.getApprovedDays}
           toggleRejectionResponseView={toggleRejectionResponseView}
           isEventBeingUpdated={isEventBeingUpdated}
           bookingDuration={bookingDuration}
@@ -179,6 +192,7 @@ const Container = Wrapped =>
           }
           createEvent={event => createEvent(event, formData)}
           updateEvent={event => updateEvent(event, formData)}
+          getTotalDays={this.getTotalDays}
         />
       );
     }
--- a/web-app/src/components/BookingModal/BookingModalForm/index.js
+++ b/web-app/src/components/BookingModal/BookingModalForm/index.js
@@ -24,6 +24,7 @@ const BookingModalForm = props => {
     booking,
     isSameDay,
     getApprovedDays,
+    getPendingDays,
     getTotalDays,
   } = props;
 
@@ -54,11 +55,20 @@ const BookingModalForm = props => {
     }
   };
 
-  const remainingDays = (getTotalDays - getApprovedDays());
-  const composeRemainingHolsErrorMessage = () => {
-    if (remainingDays <= 0 || bookingDuration > remainingDays) { 
-      return true;
-    }
+  const remainingDays = (getTotalDays - getApprovedDays);
+  const pendingDays = getPendingDays;
+  const availableDays = (remainingDays - pendingDays);
+  const showRemainingHolsAlertMessage = () => {
+    return (remainingDays === 0 || bookingDuration > remainingDays) || (availableDays < bookingDuration);     
+  };
+
+  const holidayAlertMessage = () => {
+    return (
+      <HolidayAlert>
+        <tag>{availableDays} Days Available</tag>
+        <p>You do not have enough remaining holidays, please contact a member of the HR team.</p>     
+      </HolidayAlert>
+    ); 
   };
 
   const composeErrorMessage = () => {
@@ -72,7 +82,7 @@ const BookingModalForm = props => {
         start
       );
       const daysNotice = calculateDaysNotice(bookingDuration);
-      return fromTodayToStartDateRequested < daysNotice && isSameDay ? (
+      return fromTodayToStartDateRequested < daysNotice ?  (
         <NoticeAlert>
           <p>
             <FontAwesomeIcon icon={faExclamationCircle} />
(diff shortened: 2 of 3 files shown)


web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +54,13 @@ const BookingModalForm = props => {
     }
   };
 
+  const remainingDays = (getTotalDays - getApprovedDays());
+  const composeRemainingHolsErrorMessage = () => {
+    if (remainingDays <= 0 || bookingDuration > remainingDays) {
```
ghost: Is the situation of less than zero ever going to apply?

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -90,77 +99,83 @@ const BookingModalForm = props => {
...
-          value={formData.updateMessage}
-          label="Reason for updating holiday:"
-          labelClass={isEventBeingUpdated ? null : 'hide'}
-          disabled={!formIsValid}
-        />
-      </Form>
+      {composeRemainingHolsErrorMessage() ?
+        <HolidayAlert>
+          <strong> {remainingDays} Days Remaining </strong>
+          <p> You do not have enough remaining holidays, please contact a member of the HR team. </p>
```
ghost: Could you remove the space at the start and end here.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -90,77 +99,83 @@ const BookingModalForm = props => {
...
-          }}
-          value={formData.updateMessage}
-          label="Reason for updating holiday:"
-          labelClass={isEventBeingUpdated ? null : 'hide'}
-          disabled={!formIsValid}
-        />
-      </Form>
+      {composeRemainingHolsErrorMessage() ?
+        <HolidayAlert>
+          <strong> {remainingDays} Days Remaining </strong>
```
ghost: If you could make this a tag rather than a strong tag then put `font-weight: bold;` in the styled component.

web-app/src/components/BookingModal/BookingModalForm/styled.js
```diff
@@ -18,3 +18,22 @@ export const NoticeAlert = styled.div`
     }
   }
 `;
+
+export const HolidayAlert = styled.div`
+margin: 15px 0;
```
ghost: The tabbing needs adjusted for the HolidayAlert block.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +54,13 @@ const BookingModalForm = props => {
     }
   };
 
+  const remainingDays = (getTotalDays - getApprovedDays());
```
bgribz: Should `getTotalDays` be called here rather than referenced? I.e. `getTotalDays() - getApprovedDays()`

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -90,77 +99,83 @@ const BookingModalForm = props => {
...
-          label="Reason for updating holiday:"
-          labelClass={isEventBeingUpdated ? null : 'hide'}
-          disabled={!formIsValid}
-        />
-      </Form>
+      {composeRemainingHolsErrorMessage() ?
+        <HolidayAlert>
+          <strong> {remainingDays} Days Remaining </strong>
+          <p> You do not have enough remaining holidays, please contact a member of the HR team. </p>     
+        </HolidayAlert> :
```
bgribz: Having this much code displayed via a ternary is a good case for moving it to its own component

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +54,13 @@ const BookingModalForm = props => {
     }
   };
 
+  const remainingDays = (getTotalDays - getApprovedDays());
+  const composeRemainingHolsErrorMessage = () => {
+    if (remainingDays <= 0 || bookingDuration > remainingDays) { 
+      return true;
```
bgribz: I think this could be named better as there's no composition going on in the function, and we can get rid of the `if` as well

```
const showRemainingHolsErrorMessage = () => {
  return (remainingDays <= 0 || bookingDuration > remainingDays)
}
```

bgribz (changes requested):

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -90,77 +99,83 @@ const BookingModalForm = props => {
...
+          <p> You do not have enough remaining holidays, please contact a member of the HR team. </p>     
+        </HolidayAlert> :
+        <Form formData={formData} formStatus={formStatus} actions={createCtas()}>
+          <Input
+            type="select"
+            htmlAttrs={{
+              name: 'eventTypeId',
+              options: renderWFH(),
+            }}
+
```
d1wilko, line 120: remove line break here?

mcromie92:
@girwin575 @bgribz made changes requested here, added some more functionality too.
Previously if the user had 3 days remaining they could request two days twice and both could be approved.
I have added functionality to take into account how many days they have pending against the remaining days - so if they have three days remaining and 2 pending they will only be allowed to request one day. This means the remaining holidays will never go below 0.
Hope this makes sense, any questions let me know :)

d1wilko (commented):
some real good code, one tiny issue i noticed apart from what `('G' || 'B') + 'arry'` mentioned

web-app/src/components/BookingModal/BookingModalForm/index.js
William-Unosquare: Can I take a moment @d1wilko to just appreciate this:
`('G' || 'B') + 'arry'`

web-app/src/components/BookingModal/BookingModalForm/container.js
```diff
@@ -72,6 +75,22 @@ const Container = Wrapped =>
...
+    };
+
+    getPendingDays = () => {
+      return getTotalDaysInEventArrayWithStatus(
+        this.props.allEvents,
+        HolidayStatus.PENDING
+      );
+    };
+    
+    getTotalDays = (this.props.userDetails.totalHolidays);
```
William-Unosquare, line 92: I don't think it's good practice to assign variables like this outside of function scope in a class. The naming of this also makes me think it's a function, especially alongside `getPendingDays` etc which are actually functions.

A far cleaner way to do this would be to destructure this value from props around line `183`.

web-app/src/components/BookingModal/BookingModalForm/container.js
```diff
@@ -179,6 +200,7 @@ const Container = Wrapped =>
           }
           createEvent={event => createEvent(event, formData)}
           updateEvent={event => updateEvent(event, formData)}
+          getTotalDays={this.getTotalDays}
```
William-Unosquare, line 203: Rename this prop to "totalDays". Same reason as above.

web-app/src/components/BookingModal/BookingModalForm/container.js
```diff
@@ -170,6 +189,8 @@ const Container = Wrapped =>
           toggleRejectionMessageView={toggleRejectionMessageView}
           capturedRejectionReponseText={capturedRejectionReponseText}
           assignRejectionResponseText={this.assignRejectionResponseText}
+          getApprovedDays={this.getApprovedDays()}
+          getPendingDays={this.getPendingDays()}
```
William-Unosquare, line 193: Rename these to "approvedDays" and "pendingDays". The function names are fine but the prop name makes these props sound like functions.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +55,22 @@ const BookingModalForm = props => {
     }
   };
 
+  const remainingDays = (getTotalDays - getApprovedDays);
+  const pendingDays = getPendingDays;
```
William-Unosquare, line 59: Won't be any need for this line if you rename the props.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +55,22 @@ const BookingModalForm = props => {
...
+  const availableDays = (remainingDays - pendingDays);
+  const showRemainingHolsAlertMessage = () => {
+    return (remainingDays === 0 || bookingDuration > remainingDays) || (availableDays < bookingDuration);
```
William-Unosquare, line 62: Is the `remainingDays === 0` part needed? I think `bookingDuration > remainingDays` will cover this saying it's not possible to book 0 day holidays.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -90,77 +109,79 @@ const BookingModalForm = props => {
...
-            placeholder: 'optional',
-          }}
-          value={formData.updateMessage}
-          label="Reason for updating holiday:"
-          labelClass={isEventBeingUpdated ? null : 'hide'}
-          disabled={!formIsValid}
-        />
-      </Form>
+      {showRemainingHolsAlertMessage() ? holidayAlertMessage() :
+        <Form formData={formData} formStatus={formStatus} actions={createCtas()}>
```
William-Unosquare, line 113: Pull the Form out into its own function.  A lot being rendered here off the ternary.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -175,6 +196,9 @@ BookingModalForm.propTypes = {
   updateEvent: PT.func.isRequired,
   booking: PT.object,
   isSameDay: PT.bool.isRequired,
+  getApprovedDays: PT.number,
+  getPendingDays: PT.number,
+  getTotalDays: PT.number,
```
William-Unosquare, line 201: .isRequired

William-Unosquare (changes requested):
Few things to change here, mostly around naming of variables.

web-app/src/components/BookingModal/BookingModalForm/index.js
```diff
@@ -52,6 +55,22 @@ const BookingModalForm = props => {
...
+  const pendingDays = getPendingDays;
+  const availableDays = (remainingDays - pendingDays);
+  const showRemainingHolsAlertMessage = () => {
+    return (remainingDays === 0 || bookingDuration > remainingDays) || (availableDays < bookingDuration);     
+  };
+
+  const holidayAlertMessage = () => {
+    return (
+      <HolidayAlert>
+        <tag>{availableDays} Days Available</tag>
```
ghost, line 68: I may be silly here but I've never seen the <tag> element before and it's not on https://www.w3schools.com/tags/ does it exist, if so what purpose does it serve?

PR #268, opened 2018-10-03
Remove node proxy cors as the .net one is in place

ghost:
This PR removes the nodeJS cors proxy from the project


eoinmcafee:
@trojanspike , This is quite an old pull request? Is it still required? Need approval? If so can you chase @BrownMark to please?

PR #278, opened 2018-11-06
Feature/employees

mccleaveryan:
- Stubbed out methods for Employees in controller and service 
- Reflecting swagger that exists for employee controller on java back end of adminCore


v2/AdminCore.WebApp/Controllers/AuthenticationController.cs
```diff
@@ -93,7 +93,7 @@ public ActionResult Login([FromBody] LoginRequestModel model)
     [HttpPost]
     [AllowAnonymous]
     [Route("register")]
-    public ActionResult Register([FromBody] RegisterEmployeeViewModel registerEmployee)
+    public ActionResult Register([FromBody] EmployeeViewModel registerEmployee)
```
BrownMark: I think this should still take the RegisterEmployeeViewModel

BrownMark (changes requested):
Good work, just one minor change.

BrownMark (approved):

PR #279, opened 2018-11-06
Feature/stub out working from home

Jamie-Unosquare:


v2/AdminCore.WebApp/Controllers/WorkingFromHomeController.cs
```diff
@@ -0,0 +1,51 @@
...
+    private IMapper _mapper;
+    private IEventService _eventService;
+
+    public WorkingFromHomeController(IEventService eventService, IMapper mapper)
+    {
+      _eventService = eventService;
+      _mapper = mapper;
+    }
+
+    [HttpGet("/")]
```
BrownMark: Don't need the route here.

BrownMark (changes requested):
Close :)

v2/AdminCore.WebApp/Controllers/WorkingFromHomeController.cs
```diff
@@ -0,0 +1,51 @@
...
+    private IEventService _eventService;
+
+    public WorkingFromHomeController(IEventService eventService, IMapper mapper)
+    {
+      _eventService = eventService;
+      _mapper = mapper;
+    }
+
+    [HttpGet]
+    [AllowAnonymous]
```
eoinmcafee, line 24: Can you remove all the allowed Anonymous please? If we want to test without auth we can just comment out the Authorize header at the top of the controller.

BrownMark (approved):

PR #297, opened 2018-11-30
Added client functionality and holiday create functionality

nmcmahon-unosquare:
Added responseMessage class which will be passed back from service layer to controller layer. It will contain either the object being passed back or an EmptyMessage object which signifies an object has not been passed back (e.g. an update would use this).

Built out functionality for the client controller. Standard CRUD endpoints (GetById, GetByName, Create, Update). Delete was omitted as it should not be possible to delete a client.

Built out the create functionality for holidays. This involved creating the EventService and implementing the splitting logic for new events. Validation added to holiday create also.

SharedController class added to contain code that will be used across all controllers to map responseMessage objects to http responses.

Minor schema changes to make DB migrations work first time.

Unit tests added for new functionality


v2/AdminCore.Constants/Enums/DatabaseStatus.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace AdminCore.Constants.Enums
+{
```
eoinmcafee: Do we still use this?

v2/AdminCore.DAL/AdminCore.DAL.csproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Http" Version="2.1.1" />
```
eoinmcafee: Is this required?

v2/AdminCore.Services/ClientService.cs
```diff
@@ -1,45 +1,107 @@
...
+      }
+      else
+      {
+        return new ResponseMessage<ClientDto>(null).WithStatus(MessageConstants.MsgStatusNoRecords);
+      }
+    }
+
+    public ResponseMessage<IList<ClientDto>> GetByClientName(string clientName)
+    {
+      var clientListReturnedFromDb = _dbContext.ClientRepository.Get(x => x.ClientName == clientName);
```
eoinmcafee: use string comparison here.  x.ClientName.equal(clientName, StringblaIgnoreculturecase)

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+
+    private void IfOnlyOneEventAddToList(IList<EventDate> eventDates, IList<EventDto> eventDto)
+    {
+      if (!eventDto.Any())
+      {
+        eventDto.Add(GetEvent(eventDates.First().EventId).Payload);
+      }
+    }
+
+    private int IfNewEventThenAddPreviousEventToList(EventDate eventDate, int lastEventId, IList<EventDto> eventDto)
```
eoinmcafee, line 170: All public methods at the top, private's at the bottom

v2/AdminCore.Common/AdminCore.Common.csproj
```diff
@@ -4,9 +4,19 @@
...
+  </ItemGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\AdminCore.Constants\AdminCore.Constants.csproj" />
     <ProjectReference Include="..\AdminCore.DTOs\AdminCore.DTOs.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Reference Include="Microsoft.AspNetCore.Mvc.Abstractions">
+      <HintPath>..\..\..\..\..\.nuget\packages\microsoft.aspnetcore.mvc.abstractions\2.1.3\lib\netstandard2.0\Microsoft.AspNetCore.Mvc.Abstractions.dll</HintPath>
```
eoinmcafee: Is this required here? Can you check if you delete it will it break anything in that package?

v2/AdminCore.WebApp/Models/Holiday/CreateHolidayViewModel.cs
```diff
@@ -1,10 +1,14 @@
...
+using System.Collections.Generic;
+using AdminCore.Constants.Enums;
 using AdminCore.DAL.Models;
 
 namespace AdminCore.WebApi.Models.Holiday
 {
   public class CreateHolidayViewModel
   {
-    public EventTypes EventType { get; set; }
+    public DateViewModel DateRange { get; set; }
```
eoinmcafee, line 10: What is this used for? If you are creating a holiday the controller would accept start and end date and then you'd pass the employee Id using the authenticated user service.

v2/AdminCore.WebApp/Models/Holiday/CancelHolidayViewModel.cs
```diff
@@ -2,5 +2,6 @@
 {
   public class CancelHolidayViewModel
   {
```
eoinmcafee, line 4: Is this something Mark approved? Personally i don't see why we bother with a view model for a single Id? I may be wrong though.

v2/AdminCore.WebApp/Controllers/SharedController.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
+using AdminCore.Common.Message;
+using AdminCore.Common.Message.Elements;
+using Microsoft.AspNetCore.Mvc;
+
+namespace AdminCore.WebApi.Controllers
+{
+  public class SharedController : ControllerBase
```
eoinmcafee: Please rename to BaseController. Best practices.

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
     private readonly IEventService _holidayEventService;
-
+    private readonly IEmployeeService _employeeService;
     private readonly IMapper _mapper;
 
-    public HolidayController(IEventService holidayEventService, IMapper mapper)
+    public HolidayController(IAuthenticatedUser authenticatedUser, IEventService holidayEventService,
+      IEmployeeService employeeService, IMapper mapper)
     {
+      _authenticatedUser = authenticatedUser;
```
eoinmcafee, line 36: Can we keep the this ordered alphabetically. Same for all classes.

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
     }
 
     [HttpGet("{holidayId}")]
     public IActionResult GetHolidayByHolidayId(int holidayId)
     {
-      return Ok();
+      var holiday = _holidayEventService.GetEvent(holidayId);
+      if (holiday.Status == MessageConstants.MsgStatusSuccess)
+      {
+        return Ok(_mapper.Map<HolidayViewModel>(holiday.Payload));
```
eoinmcafee: There's now inconsistency between how you are doing your returns. Niall is using the base method and Jamie is using this. Can you agree on one approach and both use it please.

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
-    public IActionResult GetHolidayByDateBetween(EventDate eventDate)
+    [HttpGet("findByDateBetween/{startDate}/{endDate}")]
+    public IActionResult GetHolidayByDateBetween(string startDate, string endDate)
     {
-      return Ok();
+      if (IsDatesValid(startDate, endDate))
+      {
+        return Ok(GetEventsBetweenDates(startDate, endDate));
+      }
+      return BadRequest("Please use a valid date format for start and/or end dates");
```
eoinmcafee, line 140: Move is constant

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
+      var holidays = _mapper.Map<ResponseMessage<IList<EventDto>>, List<HolidayViewModel>>(holidaysDtos);
+
+      return Ok(holidays);
+    }
+
+    protected bool ValidateDate(string date)
+    {
+      try
+      {
+        DateTime dt = DateTime.Parse(date);
```
eoinmcafee, line 158: Very valid way of checking parsing a date. c# 6 has a new way which is pretty nice. 
var dateToParse = Date.parse(date, var out dateToParse);
if (dateToParse != null)
 // etc etc

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
+
+    private IActionResult CreateResponseBasedOnErrors(IList<string> errors, CreateHolidayViewModel createHolidayViewModel)
+    {
+      return errors.Any() ? CreateBadRequestResponseWithListOfErrors(errors) : CreateOkResponseWithNewEvent(createHolidayViewModel);
+    }
+
+    private IActionResult CreateOkResponseWithNewEvent(CreateHolidayViewModel createHolidayViewModel)
+    {
+      var eventDto = _mapper.Map<EventDto>(createHolidayViewModel);
+      var eventDateDto = _mapper.Map<EventDateDto>(createHolidayViewModel.DateRange);
```
eoinmcafee: You shouldn't have to do this. Automapper should map the date ranges automatically if they have the same name in the eventDto and the mapping exists in the mapper

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+          SplitEventIfFallsOnAWeekend(nextEvent, nextEvent.EventDates.Last().EndDate, eventDates);
+          break;
+        }
+      }
+
+      if (eventDto.EventDates.Last().EndDate == originalEndDate)
+      {
+        EventDateDto eventDateDto = new EventDateDto
+        {
+          StartDate = eventDto.EventDates.First().StartDate,
```
eoinmcafee: First isn't safe to use. If you want to use first you should also use null propaganda. eventDto.EventDates.First?.StartDate

v2/AdminCore.Common/Message/Elements/AdditionalInfoItem.cs
```diff
@@ -1,37 +1,35 @@
...
-        public AdditionalInfoItem(string itemType, string description, string additionalInfo )
-        {
-            ItemType = String.IsNullOrEmpty(itemType) ? "" : itemType;
-            Description = String.IsNullOrEmpty(description) ? "" : description;
-            AdditionalInfo = String.IsNullOrEmpty(additionalInfo) ? "" : additionalInfo;
-
-        }
+  public class AdditionalInfoItem
+  {
+    public string ItemType { get; set; } = "";
```
KurtisTriverus: Why initialize it to an empty string here?

v2/AdminCore.Common/Message/Elements/AdditionalInfoItem.cs
```diff
@@ -1,37 +1,35 @@
...
+    /// <summary>
+    /// Constructor assigns all values for the AdditionalInfoItem.
+    /// </summary>
+    /// <param name="itemType"></param>
+    /// <param name="description"></param>
+    /// <param name="additionalInfo"></param>
+    public AdditionalInfoItem(string itemType, string description, string additionalInfo)
+    {
+      ItemType = String.IsNullOrEmpty(itemType) ? "" : itemType;
+      Description = String.IsNullOrEmpty(description) ? "" : description;
```
KurtisTriverus: replace "" with string.Empty

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    public ResponseMessage<IList<EventDto>> GetByDateBetween(DateTime startDate, DateTime endDate, EventTypes eventType)
+    {
+      ResponseMessage<IList<EventDto>> responseMessage = new ResponseMessage<IList<EventDto>>(null);
+      EventDateDto eventDateDto = new EventDateDto
+      {
+        StartDate = startDate,
+        EndDate = endDate
+      };
```
KurtisTriverus: Do you need this, seems like you're only using it in your Get, why not just use:

databaseContext.EventDatesRepository.Get(x => x.StartDate >= startDate
                                                                              && x.EndDate <= endDate
                                                                              && x.Event.EventTypeId ==
                                                                              (int)EventTypes.AnnualLeave);

v2/AdminCore.Services/EventService.cs
KurtisTriverus: I would even use FirstOrDefault as it won't throw an InvalidOperationException.

eventDto.EventDates.FirstOrDefault?.StartDate

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+      for (var day = from.Date; day.Date <= thru.Date; day = day.AddDays(1))
+        yield return day;
+    }
+
+    private EventDto MapEventDto(EventDto priorEvent, DateTime originalEndDate, DateTime eventDate, int plusDays)
+    {
+      EventDto nextEvent = _mapper.Map<EventDto>(priorEvent);
+      nextEvent.DateCreated = DateTime.Now;
+      nextEvent.EventDates.First().StartDate = eventDate.AddDays(plusDays);
+      nextEvent.EventDates.Last().EndDate = originalEndDate;
```
KurtisTriverus, line 213: Use FirstOrDefault/LastOrDefault

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+      if (existingEvent != null)
+      {
+        Event eventToReject = existingEvent;
+        eventToReject.EventStatus.EventStatusId = (int)EventStatuses.Rejected;
+        Employee employee = GetEmployeeFromEmployeeId(employeeId);
+        if (employee != null)
+        {
+          _mapper.Map(eventId, existingEvent);
+          Save(eventToReject);
+          responses.Add("Successfully rejected event " + eventId + " for employee: " + employeeId);
```
KurtisTriverus: Use string interpolation: $"Successfully rejected event {eventId} for employee: {employeeId}"

v2/AdminCore.Services/EventService.cs
```diff
@@ -0,0 +1,404 @@
...
+          responseMessage.Status = MessageConstants.MsgStatusSuccess;
+          responseMessage.Payload = responses;
+          return responseMessage;
+        }
+        responses.Add("No Employee found with an ID of: " + employeeId);
+      }
+      else
+      {
+        responses.Add("No Event found with an ID of: " + eventId);
```
KurtisTriverus: string interpolation

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
+        rejectHoliday.EventId,
+        rejectHoliday.Message,
+        int.Parse(_authenticatedUser.RetrieveUserId()));
+      return Ok(responses);
     }
 
-    [HttpGet("findByDateBetween/{eventDate}")]
-    public IActionResult GetHolidayByDateBetween(EventDate eventDate)
+    [HttpGet("findByDateBetween/{startDate}/{endDate}")]
+    public IActionResult GetHolidayByDateBetween(string startDate, string endDate)
```
KurtisTriverus, line 134: You guys should look at Parameter Binding: https://docs.microsoft.com/en-us/aspnet/web-api/overview/formats-and-model-binding/parameter-binding-in-aspnet-web-api

v2/AdminCore.Common/Interfaces/IEmployeeService.cs
```diff
@@ -27,7 +27,7 @@ public interface IEmployeeService
 
     void Delete(int employeeId);
 
-    IList<EmployeeDto> Get(int employeeId);
+    EmployeeDto Get(int employeeId);
 
     IList<EmployeeDto> GetByForenameAndSurname(string forename, string surname);
```
unosquare-terrykernan, line 32: should this be wrapped in a ResponseMessage<EmployeeDto>, so it's like the other interfaces?

v2/AdminCore.DAL/AdminCore.DAL.csproj
```diff
@@ -5,6 +5,7 @@
...
     <PackageReference Include="Microsoft.EntityFrameworkCore" Version="2.1.4" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.SqlServer" Version="2.1.4" />
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="2.1.2" />
```
unosquare-terrykernan: do we support both types of databases?

v2/AdminCore.WebApp/Controllers/HolidayController.cs
```diff
@@ -22,74 +27,184 @@ namespace AdminCore.WebApi.Controllers
...
       _holidayEventService = holidayEventService;
+      _employeeService = employeeService;
       _mapper = mapper;
     }
 
     [HttpGet]
     public IActionResult GetAllHolidays()
     {
-      return Ok();
+      var holidaysDtos = _holidayEventService.GetByType(EventTypes.AnnualLeave);
```
unosquare-terrykernan: this looks like it might return quite a lot of data, is this really what is required?

v2/AdminCore.WebApp/Controllers/HolidayController.cs
unosquare-terrykernan: DateTime dateTime = null;
        if (DateTime.TryParse("2019-01-01", out dateTime))
        {
            Console.WriteLine(dateTime);
        }

this way if it throws an exception, you can still catch it.

v2/AdminCore.WebApp/Models/Holiday/CancelHolidayViewModel.cs
unosquare-terrykernan: If this is for approving holidays, it surely needs to know who approved it, when it was approved, is that taken care of by the JWT and the current date within the save method?

unosquare-terrykernan: a message might also be required

unosquare-terrykernan (changes requested):

eoinmcafee (dismissed):

v2/AdminCore.DAL/AdminCore.DAL.csproj
eoinmcafee: Shouldn't have been added.

PR #306, opened 2018-12-06
Added Validation for Creating Holiday

Jamie-Unosquare:
Unit tests not yet added, blocked by: https://github.com/UnosquareBelfast/AdminCore/pull/303


v2/AdminCore.Services/EventService.cs
```diff
@@ -174,5 +174,39 @@ private Employee GetEmployeeFromEmployeeId(int employeeId)
       var employee = DatabaseContext.EmployeeRepository.GetSingle(x => x.EmployeeId == employeeId);
       return employee;
     }
+
+    public void IsHolidayValid(int employeeId, EventDateDto eventDates, bool modelIsHalfDay)
+    {
+      if (IsHolidayDatesAlreadyBooked(employeeId, eventDates))
+        throw new Exception("Already Booked");
```
eoinmcafee, line 230: Not sure if we can to throw an exception? Just build up a list of messages and return to front end

v2/AdminCore.Services/EventService.cs
```diff
@@ -174,5 +174,39 @@ private Employee GetEmployeeFromEmployeeId(int employeeId)
...
+      if (!IsDateRangeLessThanTotalHolidaysRemaining(employeeId, eventDates))
+        throw new Exception("Not enough holidays remaining.");
+
+      if (modelIsHalfDay)
+        throw new Exception("Holiday booked is a half day.");
+    }
+
+    private bool IsDateRangeLessThanTotalHolidaysRemaining(int employeeId, EventDateDto eventDates)
+    {
+      var employee = GetEmployeeFromEmployeeId(employeeId);
```
eoinmcafee, line 241: This logic needs to take into consideration the events booked for a year and also if there are any half days.

Jamie-Unosquare: What do you mean the events booked for a year? Will the total holidays not get updated at the start of each year anyway?

Also for half days should isn't this checked in the       if (modelIsHalfDay) check? Or do we need to implement a -0.5 to the holidays?

eoinmcafee (approved):

PR #316, opened 2018-12-10
refactoring events

eoinmcafee:


v2/AdminCore.Services/EventService.cs
```diff
@@ -42,7 +45,15 @@ public IList<EventDto> GetByDateBetween(DateTime startDate, DateTime endDate)
 
     public IList<EventDto> GetEventsByEmployeeId(int employeeId)
     {
-      var events = DatabaseContext.EventRepository.Get(x => x.Employee.EmployeeId == employeeId);
+      var startDate = _dateService.GetStartOfYearDate();
```
Jamie-Unosquare: Can we name these something like "yearStartDate", other methods use startDate and endDate in here and it might get confusing differentiating between them.

nmcmahon-unosquare (approved):

PR #317, opened 2018-12-10
Made a start on Dashboard functionality.

nmcmahon-unosquare:
- Implemented get logged in employee's events for a given month.
- Add service to Dependency injection.
- Added functionality to get logged in user from Employee Credentials
- Changed EventService to Event object is populated with Event dates.

### DIFF
--- a/v2/AdminCore.Common/Interfaces/IContractService.cs
+++ b/v2/AdminCore.Common/Interfaces/IContractService.cs
@@ -1,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using AdminCore.DTOs;
+
+namespace AdminCore.Common.Interfaces
+{
+  public interface IContractService
+  {
+    ContractDto GetContractById(int contractId);
+
+    IList<ContractDto> GetContractByEmployeeId(int employeeId);
+
+    IList<ContractDto> GetContractByTeamId(int teamId);
+
+    IList<ContractDto> GetContractByEmployeeIdAndTeamId(int employeeId, int teamId);
+
+    void SaveContract(ContractDto contractToBeSaved);
+
+    void DeleteContract(int contractId);
+  }
+}
--- a/v2/AdminCore.DTOs/ContractDto.cs
+++ b/v2/AdminCore.DTOs/ContractDto.cs
@@ -1,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace AdminCore.DTOs
+{
+  public class ContractDto
+  {
+    public int ContractId { get; set; }
+
+    public int TeamId { get; set; }
+
+    public int EmployeeId { get; set; }
+
+    public DateTime StartDate { get; set; }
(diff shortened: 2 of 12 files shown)


v2/AdminCore.Services/DashboardService.cs
```diff
@@ -3,19 +3,32 @@
...
     {
       throw new NotImplementedException();
     }
 
-    public IList<EventDto> GetEmployeeEvents(int employeeId, DateTime date)
+    public IList<EventDto> GetEmployeeEventsForAGivenMonth(int employeeId, DateTime date)
     {
-      throw new NotImplementedException();
+      var eventsForEmployee = _eventService.GetEventsByEmployeeId(employeeId);
+      return eventsForEmployee.Where(employeeEvent => EventHappensThisMonth(employeeEvent, date)).ToList();
```
BrownMark: So here we’re Retrieving all events for an employee and then delimiting in memory, it should be done at the database level

v2/AdminCore.Services/DashboardService.cs
```diff
@@ -36,5 +49,26 @@ public IList<EventDto> GetTeamDashboardEvents(int employeeId, DateTime date)
...
     }
+
+    private static bool EventHappensThisMonth(EventDto employeeEvent, DateTime date)
+    {
+      return employeeEvent.EventDates.Any(eventDate => DateRangeFallsWithinMonth(eventDate.StartDate, eventDate.EndDate, date));
+    }
+
+    private static bool DateRangeFallsWithinMonth(DateTime startDate, DateTime endDate, DateTime givenDate)
+    {
+      var date = startDate.AddDays(0);
```
BrownMark: Let’s grab some time of this, I know a better approach.

BrownMark (changes requested):

v2/AdminCore.Common/Interfaces/IDashboardService.cs
```diff
@@ -9,7 +9,7 @@ public interface IDashboardService
   {
     IList<EventDto> GetEmployeeDashboardEvents(int employeeId, DateTime date);
 
-    IList<EventDto> GetEmployeeEvents(int employeeId, DateTime date);
+    IList<EventDto> GetEmployeeEventsForAGivenMonth(int employeeId, DateTime date);
```
BrownMark: I'm not so sure about these naming conventions, perhaps a bit too long winded. Can we do something like 'GetEmployeeEventsForMonth'

BrownMark (changes requested):

v2/AdminCore.Services/DashboardService.cs
nmcmahon-unosquare: Fixed

v2/AdminCore.Common/Interfaces/IDashboardService.cs
nmcmahon-unosquare: Changed to GetEmployeeEventsForMonth

PR #320, opened 2018-12-10
Contract functionality implement

nmcmahon-unosquare:
- Service implemented
- Controller implemented
- CRUD endpoints
- View Models
- Dependency injection added
- Service and controller tests.
- Mappings

### DIFF
--- a/v2/AdminCore.WebApp/Models/Contract/ContractViewModel.cs
+++ b/v2/AdminCore.WebApp/Models/Contract/ContractViewModel.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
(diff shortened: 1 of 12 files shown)


v2/AdminCore.WebApp/Models/Contract/ContractViewModel.cs
```diff
@@ -0,0 +1,20 @@
...
+  {
+    public int ContractId { get; set; }
+
+    public int TeamId { get; set; }
+
+    public int EmployeeId { get; set; }
+
+    public DateTime StartDate { get; set; }
+
+    public DateTime EndDate { get; set; }
```
eoinmcafee, line 18: Contracts can be open ended

eoinmcafee (approved):

PR #323, opened 2018-12-12
Updated the EventDto names and Queries

Jamie-Unosquare:
Updated the EventDto names and added to the queries to return proper …

### DIFF
--- a/v2/AdminCore.DTOs/Event/EventDto.cs
+++ b/v2/AdminCore.DTOs/Event/EventDto.cs
@@ -8,27 +8,27 @@ namespace AdminCore.DTOs.Event
   public class EventDto
   {
     public int EventId { get; set; }
-    
+
     public int EmployeeId { get; set; }
-    
-    public EmployeeDto EmployeeDto { get; set; }
+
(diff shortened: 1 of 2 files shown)


v2/AdminCore.DTOs/Event/EventDto.cs
```diff
@@ -8,27 +8,27 @@ namespace AdminCore.DTOs.Event
...
     public int EventTypeId { get; set; }
-    
-    public EventTypeDto EventTypeDto { get; set; }
-    
-    public ICollection<EventMessageDto> EventMessageDtos { get; set; }
-    
+
+    public EventTypeDto EventType { get; set; }
+
+    public ICollection<EventMessageDto> EventMessage { get; set; }
```
BrownMark: Very minor, but EventMessages here please 👍

BrownMark (changes requested):

v2/AdminCore.DTOs/Event/EventDto.cs
Jamie-Unosquare: So close

BrownMark (approved):

PR #325, opened 2018-12-12
Logic for Half Days Implemented

Jamie-Unosquare:
Updated ints to doubles
Checks for single days that are also half days
Added check for if there are enough holidays before creating or updating event
Updated Tests
Updated Schema

*EventServiceTests > Splitting logic tests will fail as there is now a dependency on database mocking

### DIFF
--- a/v2/AdminCore.Common/Interfaces/IEventService.cs
+++ b/v2/AdminCore.Common/Interfaces/IEventService.cs
@@ -25,7 +25,7 @@ namespace AdminCore.Common.Interfaces
 
     EventDto CreateEvent(int employeeId, EventDateDto dates, EventTypes eventTypes);
 
-    void UpdateEvent(EventDateDto eventDateDto);
+    void UpdateEvent(int employeeId, EventDateDto eventDateDto);
 
     void UpdateEventStatus(int eventId, EventStatuses status);
 
--- a/v2/AdminCore.Services/EventService.cs
+++ b/v2/AdminCore.Services/EventService.cs
@@ -156,7 +156,7 @@ namespace AdminCore.Services
         SplitEventIfFallsOnAWeekend(newEvent, dates.EndDate, dates.StartDate);
       }
 
-      if (DoesEmployeeHaveEnoughHolidays(employeeId, newEvent))
+      if (EmployeeHasEnoughHolidays(employeeId, newEvent))
       {
         DatabaseContext.EventRepository.Insert(newEvent);
         DatabaseContext.SaveChanges();
@@ -181,7 +181,7 @@ namespace AdminCore.Services
           SplitEventIfFallsOnAWeekend(eventToUpdate, eventDateDto.EndDate, eventDateDto.StartDate);
         }
 
-        if (DoesEmployeeHaveEnoughHolidays(employeeId, eventToUpdate))
+        if (EmployeeHasEnoughHolidays(employeeId, eventToUpdate))
         {
           eventToUpdate.LastModified = _dateService.GetCurrentDateTime();
           DatabaseContext.SaveChanges();
@@ -334,7 +334,7 @@ namespace AdminCore.Services
       return true;
     }
 
-    private bool DoesEmployeeHaveEnoughHolidays(int employeeId, Event newEvent)
+    private bool EmployeeHasEnoughHolidays(int employeeId, Event newEvent)
     {
       return GetHolidayStatsForUser(employeeId).AvailableHolidays >= GetDaysInEvent(newEvent.EventDates);
     }
--- a/v2/AdminCore.Services/EventService.cs
+++ b/v2/AdminCore.Services/EventService.cs
@@ -138,31 +138,11 @@ namespace AdminCore.Services
 
     public EventDto CreateEvent(int employeeId, EventDateDto dates, EventTypes eventTypes)
     {
-      var newEvent = new Event
-      {
-        DateCreated = DateTime.Now,
-        EmployeeId = employeeId,
-        EventStatusId = (int)EventStatuses.AwaitingApproval,
-        EventTypeId = (int)EventTypes.AnnualLeave,
-        EventDates = new List<EventDate>(),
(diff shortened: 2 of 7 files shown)


v2/AdminCore.Services/EventService.cs
```diff
@@ -147,23 +147,49 @@ public EventDto CreateEvent(int employeeId, EventDateDto dates, EventTypes event
...
+        if (IsHalfDay(eventDateDto))
+        {
+          eventToUpdate.EventDates.Add(_mapper.Map<EventDate>(eventDateDto));
+        }
+        else
+        {
+          SplitEventIfFallsOnAWeekend(eventToUpdate, eventDateDto.EndDate, eventDateDto.StartDate);
+        }
+
+        if (DoesEmployeeHaveEnoughHolidays(employeeId, eventToUpdate))
```
nmcmahon-unosquare: Personal preference - but If(EmployeeHasEnoughHolidays) reads better than if(DoesEmployeeHaveEnoughHolidays).
```suggestion
        if (EmployeeHasEnoughHolidays(employeeId, eventToUpdate))
```

v2/AdminCore.Services/EventService.cs
```diff
@@ -200,7 +226,12 @@ private int GetHolidaysByEmployeeAndStatus(EventStatuses eventStatus, int employ
                                                             x => x.Employee,
                                                             x => x.EventType,
                                                             x => x.EventStatus);
-      return GetHolidaysFromReturnedEvents(events);
+      double countHolidays = 0;
```
nmcmahon-unosquare: Method looks a bit bulky. Extract 229 - 234 into own method (maybe called GetNumberOfDaysFromEvents)

v2/AdminCore.Services/EventService.cs
```diff
@@ -302,5 +333,38 @@ private bool IsHolidayDatesAlreadyBooked(int employeeId, EventDateDto eventDates
...
+    {
+      return GetHolidayStatsForUser(employeeId).AvailableHolidays >= GetDaysInEvent(newEvent.EventDates);
+    }
+
+    private double GetDaysInEvent(ICollection<EventDate> newEventEventDates)
+    {
+      double totalDays = 0;
+      foreach (var eventDate in newEventEventDates)
+      {
+        if (IsSameDay(eventDate) && eventDate.IsHalfDay)
```
nmcmahon-unosquare: Extract 347 - 354 into own method. Something like this
```suggestion
        totalDays += GetDaysFromEventDate(eventDate)
```

v2/AdminCore.Services/EventService.cs
```diff
@@ -147,23 +147,49 @@ public EventDto CreateEvent(int employeeId, EventDateDto dates, EventTypes event
...
+      {
+        DatabaseContext.EventRepository.Insert(newEvent);
+        DatabaseContext.SaveChanges();
+        return _mapper.Map<EventDto>(newEvent);
+      }
+      throw new Exception("Not enough holidays to book");
     }
 
-    public void UpdateEvent(EventDateDto eventDateDto)
+    public void UpdateEvent(int employeeId, EventDateDto eventDateDto)
```
nmcmahon-unosquare, line 150: Method is 22 lines long with some nesting and if ladders. Consider breaking it down into smaller chunks

v2/AdminCore.Services/EventService.cs
```diff
@@ -147,23 +147,49 @@ public EventDto CreateEvent(int employeeId, EventDateDto dates, EventTypes event
         EventDates = new List<EventDate>(),
         LastModified = _dateService.GetCurrentDateTime()
       };
+      if (IsHalfDay(dates))
```
nmcmahon-unosquare: Same with this method, make these if-else blocks their own methods

nmcmahon-unosquare (changes requested):

v2/AdminCore.Services/EventService.cs
```diff
@@ -302,5 +284,96 @@ private bool IsHolidayDatesAlreadyBooked(int employeeId, EventDateDto eventDates
...
+    {
+      return GetHolidayStatsForUser(employeeId).AvailableHolidays >= GetDaysInEvent(newEvent.EventDates);
+    }
+
+    private double GetDaysInEvent(ICollection<EventDate> newEventEventDates)
+    {
+      double totalDays = 0;
+      foreach (var eventDate in newEventEventDates)
+      {
+        totalDays = GetDaysFromEventDate(eventDate, totalDays);
```
nmcmahon-unosquare, line 301:
```suggestion
        totalDays += GetDaysFromEventDate(eventDate, totalDays);
```

Jamie-Unosquare: It passes in the variable in the method so += would be wrong. It would incriment the final result by the number of days

nmcmahon-unosquare (dismissed):

v2/AdminCore.Common/Interfaces/IEventService.cs
```diff
@@ -25,7 +25,7 @@ public interface IEventService
 
     EventDto CreateEvent(EventDateDto dates, EventTypes eventTypes);
 
-    void UpdateEvent(EventDateDto eventDateDto);
+    void UpdateEvent(int employeeId, EventDateDto eventDateDto);
```
unosquare-terrykernan, line 28: do we need the employeeId here? doesn't eventDateDto have an EventDTO with an employeeId?

v2/AdminCore.Services/EventService.cs
```diff
@@ -190,7 +193,8 @@ public HolidayStatsDto GetHolidayStatsForUser()
                   (startDate > x.StartDate && endDate > x.EndDate);
```
unosquare-terrykernan, line 193: this seems a bit odd, it looks like it's checking for any start date (past or future) with an end date in the future?

unosquare-terrykernan (approved):

PR #326, opened 2018-12-14
Changed GetAnnualLeave to GetEvent using EventType from controllers

Jamie-Unosquare:


v2/AdminCore.Common/Interfaces/IEventService.cs
```diff
@@ -7,7 +7,7 @@ namespace AdminCore.Common.Interfaces
 {
   public interface IEventService
   {
-    IList<EventDto> GetAnnualLeaveByEmployee(int employeeId);
+    IList<EventDto> GetEventByEmployee(int employeeId, EventTypes eventType);
```
nmcmahon-unosquare, line 10: Now that you've changed this at a service level, is there not a corresponding change at the controller level?

Jamie-Unosquare: It's not currently being used by holiday or WFH controllers

UnoJonnyKhattak (approved):

PR #330, opened 2019-01-02
Code Cleanup for event service

Jamie-Unosquare:
Changed variables to use var
Extracted methods for cleaner code
Inverted some if statements and using ternary operator

### DIFF
--- a/v2/AdminCore.Services/EventService.cs
+++ b/v2/AdminCore.Services/EventService.cs
@@ -116,7 +116,7 @@ namespace AdminCore.Services
         var employee = GetEmployeeFromEmployeeId();
         if (employee != null)
         {
-          var eventMessage = BuildRejectEventMessage(message);
+          var eventMessage = BuildEventMessage(message, MessageType.Rejected);
 
           eventToReject.EventMessages.Add(eventMessage);
           DatabaseContext.SaveChanges();
@@ -223,7 +223,7 @@ namespace AdminCore.Services
       {
         if (day.DayOfWeek == DayOfWeek.Saturday)
         {
-          SetEventDateToPreviousFriday(newEvent, startDate, day);
+          SetEndDateToPreviousDay(newEvent, startDate, day);
           var nextStartDate = day.AddDays(2);
           SplitEventIfFallsOnAWeekend(newEvent, originalEndDate, nextStartDate);
           break;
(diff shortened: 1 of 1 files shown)


v2/AdminCore.Services/EventService.cs
```diff
@@ -355,14 +345,10 @@ private void ValidateRemainingHolidaysAndUpdate(Event eventToUpdate)
...
     {
-      if (EmployeeHasEnoughHolidays(newEvent))
-      {
-        DatabaseContext.EventRepository.Insert(newEvent);
-        DatabaseContext.SaveChanges();
-        return _mapper.Map<EventDto>(newEvent);
-      }
-
-      throw new Exception("Not enough holidays to book");
+      if (!EmployeeHasEnoughHolidays(newEvent)) throw new Exception("Not enough holidays to book");
```
BrownMark: Try not to keep if statements on 1 line like that. Much cleaner to read like:

if (!EmployeeHasEnoughHolidays(newEvent)) 
{
    throw new Exception("Not enough holidays to book");
}

v2/AdminCore.Services/EventService.cs
```diff
@@ -355,14 +345,10 @@ private void ValidateRemainingHolidaysAndUpdate(Event eventToUpdate)
...
+      DatabaseContext.EventRepository.Insert(newEvent);
```
BrownMark: If you return the result from this it means the new ID will be returned.

v2/AdminCore.Services/EventService.cs
```diff
@@ -378,5 +364,17 @@ private Event BuildNewEvent(int employeeId)
       };
       return newEvent;
     }
+
+    private EventMessage BuildRejectEventMessage(string message)
```
BrownMark: Seems like this could be reused for different event message types, could pass it as a parameter?

v2/AdminCore.Services/EventService.cs
```diff
@@ -251,6 +240,16 @@ private void SplitEventIfFallsOnAWeekend(Event newEvent, DateTime originalEndDat
       newEvent.EventDates.Add(lastDate);
     }
 
+    private static void SetEventDateToPreviousFriday(Event newEvent, DateTime startDate, DateTime day)
+    {
+      newEvent.EventDates.Add(new EventDate()
```
BrownMark, line 245: Does this actually do what the method name indicates?

BrownMark (changes requested):

BrownMark (approved):

PR #349, opened 2019-01-16
Task/restore authenticated user

nmcmahon-unosquare:


v2/AdminCore.Services/AuthenticatedUser.cs
```diff
@@ -0,0 +1,34 @@
...
+
+    public AuthenticatedUser(IHttpContextAccessor httpContextAccessor, IEmployeeService employeeService)
+    {
+      _httpContentAccessor = httpContextAccessor;
+      _employeeService = employeeService;
+    }
+
+    public EmployeeDto RetrieveLoggedInUser()
+    {
+      var userDetails = GetLoggedInUserDetails();
```
BrownMark, line 24: Might be worth doing 2 things here:
1. Checking does that key exist in the dictionary.
2. store the key itself in a constants file (along with any other details we may query from the dictionary).

BrownMark (changes requested):
Looks good, just minor suggestion

v2/AdminCore.Services/AuthenticatedUser.cs
nmcmahon-unosquare: Done

nmcmahon-unosquare:
Closing as new azure approach being used.

PR #356, opened 2019-01-23
Task/add total holidays logic

Jamie-Unosquare:


v2/AdminCore.Services/EmployeeService.cs
```diff
@@ -26,14 +28,27 @@ public string Create(EmployeeDto newEmployeeDto)
...
+    private void AddPublicHolidays(Employee employee)
+    {
+      /*      if (IsNorthernIrishEmployee(employee))
+            {*/
+      AddNorthernIrishPublicHolidays(employee);
+      /*      }
+            else
+            {
+              AddMexicanPublicHolidays(employee);
+            }*/
```
nmcmahon-unosquare: Remove commented out code and replace with TODO

v2/AdminCore.Services/EmployeeService.cs
```diff
@@ -122,5 +133,123 @@ private static string EncodePasswordToBase64(string password)
...
+
+          case (int)Months.December:
+            holidays = 3;
+            break;
+        }
+      }
+
+      holidays += 3;
+      return holidays;
+    }
```
nmcmahon-unosquare, line 197: This is fine for MVP given our time constraints, but add a TODO to migrate this to a DB table later

v2/AdminCore.Services/EmployeeService.cs
```diff
@@ -122,5 +133,123 @@ private static string EncodePasswordToBase64(string password)
...
+        IsHalfDay = false
+      };
+
+      eventService.CreateEvent(christmasDay, EventTypes.PublicHoliday,
+        employee);
+      eventService.CreateEvent(newYearsDay, EventTypes.PublicHoliday,
+        employee);
+      eventService.CreateEvent(lastMondayOfMay, EventTypes.PublicHoliday,
+        employee);
+    }
```
nmcmahon-unosquare, line 245: Logic looks fine, bit of a chonker method though, extract some of these object initialisations into a builder method similar to how you've done for last monday in may. Maybe consider using a loop for the create event bit at the end if it makes sense

v2/AdminCore.Services/EmployeeService.cs
```diff
@@ -122,5 +133,123 @@ private static string EncodePasswordToBase64(string password)
...
+      var daysInMonth = DateTime.DaysInMonth(DateTime.Now.Year, (int)Months.May);
+
+      for (var day = daysInMonth; day > 0; day--)
+      {
+        var currentDateTime = new DateTime(DateTime.Now.Year, (int)Months.May, day);
+        if (currentDateTime.DayOfWeek == DayOfWeek.Monday)
+          return currentDateTime.Day;
+      }
+
+      throw new Exception("Error finding last Monday in May.");
```
nmcmahon-unosquare, line 258: When would this be thrown?

v2/AdminCore.Services/EventService.cs
```diff
@@ -187,6 +198,18 @@ public void IsEventValid(EventDateDto eventDates, bool modelIsHalfDay, int emplo
...
+    {
+      var publicHolidayEventTypeId = (int)EventTypes.PublicHoliday;
+      var publicHolidays = DatabaseContext.EventRepository.Get(x => x.EventTypeId == publicHolidayEventTypeId);
+      publicHolidays.Select(c =>
+      {
+        c.EventStatusId = (int)EventStatuses.Approved;
+        return c;
+      }).ToList();
+      DatabaseContext.SaveChanges();
+    }
```
nmcmahon-unosquare: We should definitely keep this method for any batch jobs we do, but not sure why it's included now. Rather than adding events and then retroactively approving them, why don't we just set event status as approved before adding?

nmcmahon-unosquare (approved):

PR #360, opened 2019-01-24
Code Clean Up

Jamie-Unosquare:
Removed unused classes
Reordered and removed unused imports
Updated namespaces and moved classes into folders
Removed a stray semi colon

### DIFF
--- a/v2/AdminCore.Services.Tests/AdminCore.Services.Tests.csproj
+++ b/v2/AdminCore.Services.Tests/AdminCore.Services.Tests.csproj
@@ -7,15 +7,17 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Remove="NewFolder\**" />
+    <EmbeddedResource Remove="NewFolder\**" />
+    <None Remove="NewFolder\**" />
+  </ItemGroup>
+
+  <ItemGroup>
     <PackageReference Include="AutoFixture" Version="4.5.1" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.8.0" />
     <PackageReference Include="NSubstitute" Version="3.1.0" />
     <PackageReference Include="xunit" Version="2.3.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.3.1" />
-  </ItemGroup>
-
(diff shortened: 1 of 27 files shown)


v2/AdminCore.Services.Tests/AdminCore.Services.Tests.csproj
```diff
@@ -6,6 +6,12 @@
     <IsPackable>false</IsPackable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Compile Remove="NewFolder\**" />
```
BrownMark, line 10: This line isn't needed. It's to exclude the folder, but you've deleted it anyway.

v2/AdminCore.Services/DashboardService.cs
```diff
@@ -214,12 +213,16 @@ private static string GetLocationFromEvent(Event @event)
       {
         case 1:
           return EmployeeLocationConstants.OnHoliday;
+
         case 2:
           return EmployeeLocationConstants.WorkingFromHome;
+
```
BrownMark, line 219: EventType enum can be used here rather than 1,2,3,4 etc

v2/AdminCore.WebApp/ErrorHandling/ErrorHandlingMiddleware.cs
```diff
@@ -1,41 +1,43 @@
...
+      var code = HttpStatusCode.InternalServerError; // 500 if unexpected
 
-    if (exception is UserNotRegisteredException) code = HttpStatusCode.Forbidden;
+      if (exception is UserNotRegisteredException) code = HttpStatusCode.Forbidden;
 
-    var result = JsonConvert.SerializeObject(new { error = exception.Message });
-    context.Response.ContentType = "application/json";
-    context.Response.StatusCode = (int)code;
-    return context.Response.WriteAsync(result);
+      var result = JsonConvert.SerializeObject(new { error = exception.Message });
```
BrownMark, line 37: We shouldn't be using JSON.net inline like this. Dependency Inversion is key here.

BrownMark (changes requested):

